Issue #7, opened 2019-02-15
Propose architecture change


elsevers:
- Think about injecting a class which implements the drawing functionality.
- Think about components as cross sections.
- Components export line/arc primitives

THOUGHTS:

XFEMM < Tool & Drawer2D & MakeSolid
   - implements abstract methods

Tool
     - open()
     - close()
     - cmd()
     - select()
     - extrude()
     - revolve()

- Drawer2D
    - makeArc()
    - makeLine()
    - select() (this is optional (?))

- Component
    - model() { for all CrossSections: 1. cs.draw(), 2. cs.select()           3. MakeSolid.make() }
    - CrossSection objects []
    - Material obj
    - MakeSolid obj
    - Location3D obj

- CrossSection (compBase)
    - draw()
    - select()
    - Drawer2D _object (just new line, new arc, select result)
    - Dimensions object
    - Location2D object (always in xy plane)

- MakeSolid (?)
    - make() {
         revolve() / extrude()
      }

elsevers:
main()

// Make CrossSections
CrossSection cs1;

// Assemble into Component
Component stator(MakeExtrusion, ......);
c.add(cs1);

stator.model();
// TODO: change model name to "make()"

npetersen2:
This issue served as idea space for new architecture. As this has been implemented in PR #8, I am closing this issue. Will open new issues for bugs and new ideas.

PR #8, opened 2019-02-17
New system architecture

npetersen2:
Initial working implementation of new proposed architecture (see #7).

This PR and merge is meant to completely replace the current `develop` branch.

Branch `develop` is ahead of this PR as base of `dev-new-arch` was `develop` prior to adding trapezoid component. That will need to be manually added to this new framework before this can completely replace current `develop` branch.

### Examples

Try examples:
- `examples/main_mn.m`
- `examples/main_tikz.m`

These use the same code to generate a Component (just arc for now), then draw this component using either MagNet or TikZ commands. MagNet will open the MagNet application view, while TikZ will generate `output.txt` file which is a functioning LaTeX source file which can be rendered to view the arc component.


npetersen2:
We need to discuss assigning locations and rotations to cross sections and components in this new architecture. I was trying to reimplement Eric's 4x arc example where the arcs are rotated 90 degrees around the center.

Currently, each cross section has a Location2D, which specifies XY plane location. Each component is a collection of cross sections, and it has one Location3D. This doesn't allow implementation of Eric's example. *How should we change the arch to support this?*

elsevers:
> We need to discuss assigning locations and rotations to cross sections and components in this new architecture. I was trying to reimplement Eric's 4x arc example where the arcs are rotated 90 degrees around the center.
> 
> Currently, each cross section has a Location2D, which specifies XY plane location. Each component is a collection of cross sections, and it has one Location3D. This doesn't allow implementation of Eric's example. _How should we change the arch to support this?_

I don't follow this problem -- it looks to me like you are able to specify the rotation of the cross-section.

However, I do notice that the `Location2D` describes it rotate property as:
```rotate_xy = [DimRadian(0),DimRadian(0)];   %Angles about global xy axes to rotate component's xy axes in radians   ```
We need to discuss / think about this. A 2D cross-section can really only rotate about one axis: the z axis. So I am curious how you are interpreting `rotate_xy` in the code.

elsevers:
Thoughts on `drawer`

- `drawer` is being passed into `Component.make()`, as opposed to being a property of `CrossSectBase`. Let's talk about if this is the approach we want to take. It is different than what I was expecting, but the more I think about it, the more sense it makes. (i.e. _why should a property of a cross-section be tied to a specific drawer?_)
- `CrossSectBase.select(obj)` should be `CrossSectBase.select(obj, drawer)` (or selector?)

npetersen2:
Update Location2D class to have a single rotation property. See Nathan's picture for details. Also, update comments in class source to explain dimensions, add simple drawing.

Create GitHub Issue for Nick to go do this. New topic branch, PR, etc.

npetersen2:
Update code to get rid of handler class inheritance. Should be able to just return obj from obj method and MATLAB will overwrite local obj... Need to test.

npetersen2:
remove CrossSection.select()

CS.draw() returns a token as local variable

TODO: need to think about how to select cross sections. Probably a function of the Tool class:
`Tool.prepareForMake(list of tokens from earlier)`
This is internal to the `Tool.makeSolid()`

Tool.makeSolid( tokensFromdrawing cross sections)

npetersen2:
@elsevers Please send me TikZ function for drawing arc from center point and edge points.

npetersen2:
@elsevers I've merged latest `develop` into this branch (basically just trapezoid cross section code). I believe this branch will merge cleanly back to `develop`. Feel free to click "merge" button!

PR #19, opened 2019-02-23
Add coordinate transformation method to location2D class.

Nick-Hemenway:
Modify trapezoid and arc cross sections to reflect changes
to location2D.
Add figure for reference for location2D.


npetersen2:
`rotate_xy` is a little unclear to me.

Can we change `Location2D` property `rotate_xy` to `theta`? This would match your PDF diagram.

elsevers:
Additional critiques:

- The properties and methods do not have the proper commenting style to be compatible with Matlab's [self-documentation features](https://www.mathworks.com/help/matlab/matlab_prog/create-help-for-classes.html). Please fix. 

- Why does `transformCoords(obj, x, y, theta)` have the argument `theta`? Based on our discussion, `theta` should be the property now labeled as `rotate_xy`. That is, the single responsibility of Location2D is to specify a complete transformation for coordinates.

- `transformCoords(obj, x, y, theta)`: I suggest that you change this to be `transformCoords(obj, xyCords)`, where `xyCords` is a matrix of coordinates (see below). This will allow for compact usage when calling the function.
``` 
xyCords = [x1, y1 
         x2, y2 
         x3, y3
         x4, y4];
``` 
Allows for:
`newCords = locObj.transformCoords(obj, [arcStartXY; arcCenterXY; arcEndXY]);`

elsevers:
Please add issue to track remaining items from this comment trail

PR #21, opened 2019-02-23
Inner rotor

Nick-Hemenway:


npetersen2:
Seems very good. Only concern I have is code readability. In cross section classes, code which calculates points and transforms them is a bit hard to follow:

From trapezoid cross section:
```MATLAB
    x = [-(w/2), -(w/2) + (h/tan(theta)), ...
          (w/2) - (h/tan(theta)), (w/2)];
            
    y = [0, h, h, 0];
            
    [x_trans, y_trans] = obj.location.transformCoords(x,y);

    point1 = [x_trans(1), y_trans(1)];
    point2 = [x_trans(2), y_trans(2)];
    point3 = [x_trans(3), y_trans(3)];
    point4 = [x_trans(4), y_trans(4)];
```
You completely separate the x from y coordinates. This is a little hard to understand when reading the code.

For example, if I want to know what the equation is for `point3`, I have to go find `x_trans` and `y_trans` separately, then parse out the matrix line to find: `point3 = [(w/2) - (h/tan(theta)), h]`. Maybe we can directly write that question, then transform on the whole `[x,y]` object? Storing `x` and `y` separately for each point seems unclear.

(This style is also done in other cross section classes)

npetersen2:
Another minor comment:
In `CrossSectInnerRotorPMStator.m`, when you validate the attributes, isn't `stator_slots` an integer? So following code should not validate `double`, but maybe `int` or equivalent?

```c
validateattributes(obj.stator_slots,{'double'},{'nonnegative','nonempty'})
```

elsevers:
- Missing method documentation
- When I run the example I get a lot of these messages (see below). Do we need to add more terminating `;`?
```
ans =

   NaN
```

elsevers:
I have created a pull request into this branch to change `stator_slots` to `Q`.

elsevers:
Please split into two pull requests:

1. a pull request we can complete right now (after you resolve merge conflicts) that does not completely implement fillets.
2. a pull request that adds the fillets.

elsevers:
`CrossSectInnerRotorPMStator.m` should be `CrossSectInnerRotorStator.m` (does not have anything to do with whether the rotor has PMs)

npetersen2:
The example works as is, but doesn't work to extrude multiple cross sections. (i.e., changing the `cs` variable to include all `stator`s). We can just merge the code as is.

npetersen2 (dismissed):
The example works as is, but doesn't work to extrude multiple cross sections. (i.e., changing the `cs` variable to include all `stator`s). We can just merge the code as is.

elsevers (approved):
Looks good. @Nick-Hemenway, please continue the improvements you mention in a new PR.

npetersen2 (approved):

PR #26, opened 2019-02-24
Update TikZ library to better support drawing arcs

npetersen2:
@Nick-Hemenway I tried your new inner rotor cross section example for drawing to TikZ. The TikZ library I had made didn't work, so I fixed it.

Only updates are to TikZ.m library, so this should merge easily into `develop`.


npetersen2:
I've changed this PR to merge the `TikZ.m` changes directly to the `develop` branch (that is the cause of the change in base branch and force-pushing etc).

Issue #13, opened 2019-02-21
Implement `makeSolid()` action


npetersen2:
This is still fuzzy to me about how to implement this... We need to iron this out.

---

The following is how the system will work(?) from top level `main.m`:

```MATLAB

% Make cross sections....
cs1 = ___;
cs2 = ___;

% Make components....
comp1 = ___cs1, cs2___;

% Draw component in tool
comp1.make(Tool t)

```

Tasks:
- remove CrossSection.select()
- CrossSection.draw() returns a token as local variable

```MATLAB
Tool.makeSolid(tokensFromDrawingCrossSections) {
  foreach token:
    % Tool specific code here that uses token to "select" the component primitives...
}
```

elsevers:
See pull request #32

elsevers:
Pull request #32 merged. New issue created to apply this to other cross-sections: #47

PR #32, opened 2019-02-25
First attempt at makeSolid

elsevers:
Implementation of makeSolid capabilities based on our discussion on Feb 18th as documented in issue #13.  This is meant to seed discussion for the meeting on 2/25/2019.

Key changes:

- This also commits a plumbing function to add extrusion capabilities for MagNet.

- There does not appear to be a way to extract information about a cross-section from the tokens generated by MagNet drawing lines and arcs. To overcome this issue, a new `CrossSectToken` class has been introduced. This class encapsulates information generated when drawing a
cross-section. There are two key pieces of information in here:

   1. `token` created by the drawer object, likely consists of a token created for each edge.
   2. `innerCord`: an x/y coordinate that lies somewhere within the surface of the cross-section.

Several things remain to be worked out:

- what object should house `location`, `material` and `dim_depth` information? Currently this is spread out between the `Component` object and the `MakeSolidBase` classes. 
- These commits do not check / convert unit type for extrusion depth or the coordinate of `CrossSectToken`
- After pull request #19 is finalized and merged, the calculation of `innerCoord` in the `CrossSectArc` and `CrossSectTrapezoid` classes needs to be updated. Currently, an approximation is used.
- Handling of materials
- Implementation of `Location3D`


elsevers:
I forgot to add:  run `main.m` to see the example create a solid 3D arc.

PR #40, opened 2019-02-26
Modify Location2D class transformCoords method

Nick-Hemenway:
change rotate_xy to theta and optional theta arg to add_theta.


elsevers:
The syntax doesn't follow our agreed upon convention. Please change `add_theta` to `addTheta`

PR #41, opened 2019-02-26
Fix cross-sections broken by Location2D.transformCoords revision

elsevers:
Merging in @ngadiyar93's pull request #4 broke the examples because it was not based on the most recent Location2D class. This pull request fixes that problem.


elsevers:
I am going to merge this right away because it restores functionality to `develop` for others who are planning to merge `develop` into their branches.

Issue #44, opened 2019-02-27
Operator overloading on dimension objects


npetersen2:
Currently adding two Dimensions results in a double. We should overload the the +/- etc (div, mult) operators to support this.

npetersen2:
I've started in on this. See branch `dim-operator-overload`, and PR #53.

See this link for all possible operands: https://www.mathworks.com/help/matlab/matlab_oop/implementing-operators-for-your-class.html#br02znk-6

Which of these do we want?

npetersen2:
We want overloaded:
- plus, minus, comp (>, <, ==, etc)
- scalar operations (div, mul, etc)

npetersen2:
In progress in PR #53.

PR #49, opened 2019-02-28
Add relative function to Location2D class

Nick-Hemenway:
I've added a `relative()` function to the `Location2D` class. You input a `DimLinear` and `DimAngular` object to the function which represents a relative movement from the current state and it returns a new location object with the relative movements included. The function is not ready to merge yet as it is waiting on the dimension objects operators to be overloaded. Using relative movements may make some assembly features easier.


npetersen2:
I will post update here when #44 is complete.

Bharat-Ramadas (approved):

elsevers (approved):

PR #53, opened 2019-03-04
Dimension operator overloading

npetersen2:


npetersen2:
This is not ready to merge. We need to discuss how I'm doing this and make sure it looks good before I go and add this functionality to all dimension classes.

npetersen2:
Make unit test area in repo for things like the example script.

npetersen2:
Still working on this. Need to work on implementing scalar multiplication and division.

I think to do scalar multiplication, I need to override the `times` operator and verify that one of the two arguments is strictly a double. If not, then throw an error. Similarly for division.

elsevers (approved):
Looks great. Let's merge for now, but create a new PR to add:

- functionality the angular dimensions as well
- implement remaining operators: multiplication and division (scalars)
- validate ability to function on arrays.

Nick-Hemenway (approved):

PR #64, opened 2019-03-16
Fix MagNet error with exampleBreadLoaf

elsevers:
Each component must have a unique name (or MagNet will throw an error, but this makes sense anyway, your model shouldn't have two identically named components). This PR simply changes the example code so that each component has a unique name.


elsevers:
This is ready to review and merge.

npetersen2 (approved):

ngadiyar93 (approved):

PR #74, opened 2019-03-22
Clone method is added for Class Component and CrossSectBase as a new feature.

horychen:
The effect: One cannot clone using the same name as the parent object.
Bonus effect: Moreover, one actually cannot clone using any name that has been used before.

Please run exampleConstructor.m to test it.

One thing worth mentioning is:
Matlab operates weird as always. 
If you keep getting error "Nice try blah blah", you must use "clear all" instead of "clear" to deeply clear your workspace, as the Class Component will take records of created objects.


elsevers:
@horychen Quick work! I am not sure about the approach taken here, but I will comment on the commit message. Please follow the guidelines here, going forward: https://github.com/Severson-Group/severson_group_git#sub_commit_msg

Also, if you are ready to have us review your code, add us all as reviewers.

horychen:
image
I apply this approach to Class CrossSectHollowCylinder as well. It works as expected.
~~However, I did realize one problem. This clone method needs to be implemented for each child class (i.e., cannot be inherited). As a matter of fact, this kind of makes sense, because there can be two objects (one is a CrossSectHollowCylinder while the other is CrossSectHollowRectange) having the same name (e.g., 'MySection1'). Apparently these two are children of base class CrossSectBase.~~

elsevers:
Hello Jiahao, I am skeptical that this is the best approach. This seems like a huge rewrite and appears to pull in a lot of framework... It is far more involved than what we were intending in our discussion.

Have you tried simply adding a constructor to the base class that takes as an input an old object? That would have no need for any `clone` function or inheriting `matlab.mixin.Heterogeneous` and `matlab.mixin.Copyable` Since the properties are protected, you can edit them from within the class (hence using the key value pair matching).

horychen:
I agree. In fact, that's my first attempt and it is what I will do in C++. However, what I found out is that matlab does not support an alternative constructor. I quote: "No. The constructors in OOP matlab are very restricted compared to other languages."

~~I will try to implement a static method instead. Something like this:
comp1 = Component('name', 'comp1', 'location2D', 'blah')
comp2 = Component.clone(comp2, 'name', 'comp1', 'location2D', 'blah')~~

Update:
I have reduce the changes in codes to a minimum (reuse codes) and in the meanwhile achieve the clone function. There will be zero coding to implement any subclass of CrossSectBase and Component.

The 7th commit is a mistake... I don't know how to properly undo this and added an 8th commit.

elsevers:
Thanks for your hard work, but this implementation is too complicated. Let's go with the approach in #78.

PR #75, opened 2019-03-27
Change add_theta to addTheta in Location2D

Nick-Hemenway:
I believe the Location2D should be fixed now but I cannot tell for sure because we need to add multiplication of a double and dimensional quantity to the operator overloading.


ngadiyar93 (dismissed):
Update of _add_theta_ to _addTheta_ looks good to me.

elsevers (approved):
Looks great to me

ngadiyar93 (approved):

PR #76, opened 2019-03-27
Add multiplication with non-dimensional scalar to operator overloads.

Nick-Hemenway:
The following situation will work:
```
a = DimMillimeter(2)
b = 2*a  %b will be a DimMillimeter object
```
The following will NOT work and will throw an error:
```
a = DimMillimeter(2)
b = a*2  % non-dimensional scalar cannot come after DimLinear object
```

### DIFF
--- a/model_obj/dimensions/DimLinear.m
+++ b/model_obj/dimensions/DimLinear.m
@@ -18,6 +18,17 @@ classdef DimLinear < DimBase
             r = feval(class(lhs), DimMillimeter(sum));
         end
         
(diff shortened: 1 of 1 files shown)


ngadiyar93 (approved):

elsevers (commented):
Looks good. But further work needed:

- Add a unit test
- Make this operation work in both directions (either `lhs` or `rhs` should be able to be a scalar, but not both)

elsevers (approved):

PR #78, opened 2019-03-27
Clone alternate implementation

elsevers:
Alternate implantation for cloning a object.


elsevers:
Not yet ready for review. @horychen will finish this PR.

elsevers:
Note that this PR replaces #74.

npetersen2 (approved):
I didn't run this, but the code looks good to me. Very clean and nice!

examples/exampleClone.m
```diff
@@ -0,0 +1,40 @@
...
+            'location', Location3D( ...
+                'anchor_xyz', DimMillimeter([0,0,0]), ...
+                'rotate_xyz', DimDegree([0,0,0]).toRadians() ...
+                ), ...
+            'dim_depth', DimMillimeter(80)) ...
+        );
+    
+comp2 = comp1.clone('comp2', 'crossSections', bread2);
+
+%% This is example of a potential bug: comp3 has the same name as comp1, and bread3 has the same name as bread1.
```
elsevers, line 37: Thanks for noting this. I don't think that this is an enforceable issue. Two components can always be created with the same name using the class constructors (not via cloning). It is the user's responsibility to monitor this. Trying to track and enforce it will cause problems.

elsevers:
I checked out the code and tried out the example. This looks good to me. I am not able to be a reviewer because I created the pull request.

**Let's get some more reviews -- I think this is ready to merge**

ngadiyar93 (approved):
I think this looks good

PR #82, opened 2019-03-28
Add division of two DimLinear operation for making a ratio

npetersen2:


npetersen2:
Note that this probably allows a user to divide two different types.... `DimInch / DimMillimeter`... Is this okay? The code casts both to LHS and RHS to same type (i.e. `DimMillimeter`) for the division, so I think this is the expected behavior...

npetersen2:
Update: I have added another unit test to show the above commented behavior (see last test). I am now confident that this is the expected behavior. Please approve this PR so I can merge to `develop`.

elsevers:
If they can divide two different types, the units should be converted to a consistent unit base before the divide operation. So, the scalar output for:
``` 
mmInch = dimMillimeter(25.4);
Inch = dimInch(1);
b = mmInch/Inch;
```

The result should be `b = 1` (double)

npetersen2:
@elsevers Yes, this is what happens. See the expected result for the last test case (this passes).

Nick-Hemenway (approved):

elsevers (approved):
I ran the tests and took a look around. Looks great.

PR #83, opened 2019-04-01
Update tests to ensure comparison is using double types

npetersen2:
This doesn't really change anything, just a sanity check. The code prior to change could be using the overloaded `-` operator, which is not desired.


npetersen2:
Note: all the tests `runTests.m` still pass. This should be a trivial merge.

horychen (approved):
Since you said it is some trivial changes...

ngadiyar93 (approved):

PR #105, opened 2020-01-03
Ansys mech

Nick-Hemenway:
This branch contains the Ansys Mechanical APDL tool using Ansys as a Server. At this point there is a working example that draws the cross section of the stator or the outer-rotor machine but the tool is not yet capable of performing structural analyses.


elsevers:
@Nick-Hemenway, can you provide instructions for how we can run your test code? (Which server and any instructions that we need to follow before running your test script)

Nick-Hemenway:
The test code works on servers 01 and 02 but not 03 because APDL is not installed there. in the `tools --> ansysMech` folder there is a word file called `AnsysMechExample` that explains how to set up ansys as a server and how to run the example code that I've included.

PR #112, opened 2020-09-24
Add crossections

### LINKED ISSUE #108: Add new crosssects
Add `crosssects` that @wchan29 and @ngadiyar93 developed for conical PM bearing modeling.

ngadiyar93:
Add crosssects that were developed for conical PM bearing modeling.


Bharat-Ramadas (dismissed):
Works fine

model_obj/crosssects/parallelogram/CrossSectParallelogram.m
```diff
@@ -0,0 +1,74 @@
...
+            [points] = obj.location.transformCoords([x',y']);
+            
+            %% Draw Parallelogram
+            [l_1] = drawer.drawLine(points(1,:),points(2,:));
+            [l_2] = drawer.drawLine(points(2,:),points(3,:));
+            [l_3] = drawer.drawLine(points(3,:),points(4,:));
+            [l_4] = drawer.drawLine(points(4,:),points(1,:));
+            
+            %compute coordinate inside the surface to extrude
+            x_coord = (l*cos(theta)+ t/sin(theta))/2;
```
elsevers, line 41: I am not sure about your `x_coord` math. Can you please explain it in a response to this comment?

elsevers (changes requested):
Nice job. I have a question about the `x_coord` calculation that I have inserted as an inline comment.

In the drawing, please make your label text consistent with other figures by italicizing your variable names.

model_obj/crosssects/parallelogram/CrossSectParallelogram.m
ngadiyar93: The array `x` has the x-coordinates of the 4 points forming the parallelogram.
`x_coord` basically is the coordinate of a point **inside** the surface to extrude. Ideally we would want it to be the mid-point.
 
Because this is a parallelogram and no two points have the same x coordinates, what @wchan29 did here was take the x_coord equal to the mid point of the left top point and the right bottom point. (I have circled the points here).

image

elsevers (approved):
Thanks @ngadiyar93. This looks good to me.

Bharat-Ramadas (approved):
Looks good

PR #127, opened 2020-10-16
Add support for MagNet tool class in Python

### LINKED ISSUE #124: Create MagNet tool class in Python
Please create a Python version of `eMach/tools/magnet/MagNet.m`. 

This will require you to create associated abstract base classes `ToolBase`, `DrawerBase`, `MakerExtrudeBase`, `MakerRevolveBase`

As you do this, think about the broader picture for how similarly we should structure the Python code base to the MATLAB code base. Consider the questions posed in issue #110.

Bharat-Ramadas:
# Description

This PR is to merge the python implementation of MagNet abstract base classes already implemented in Matlab. 

# Supported Abstract Base Classes
## `ToolBase`
This base class, if inherited, necessitates the inclusion of a `open` and `close` methods in the Child Class.
## `DrawerBase`
This base class, if inherited, necessitates the inclusion of `drawLine`, `drawArc` and `select` methods in the Child Class.
## `MakerExtrudeBase`
This base class, if inherited, necessitates the inclusion of  `extrude` method in the Child Class.
## `MakerRevolveBase`
This base class, if inherited, necessitates the inclusion of  `revolve` method in the Child Class.
## `MakerBase`
This base class, if inherited, necessitates the inclusion of `prepareSection` method in the Child Class.

# Supported Tool Classes
## `MagNet`
The `MagNet` class shows how the abstract base classes can be inherited by another class. MagNet inherits all the classes mentions above and has therefore has all the methods mentioned above as well.

# Known Bugs

1. The current implementation does not check the attributes of the arguments passed to the class methods except for `drawLine`. This will be rectified in future updates.
2.  `close` and `select` methods have not been implemented in `MagNet`

# Related Issues

 #124


elsevers (changes requested):
Good first attempt at this and your work led to an interesting discussion in our meeting.

Here are the outcomes for revision as I recall:
- [x]  Use Python style for naming, but keep the meaning of the names as similar to the MATLAB code as possible. See [here](https://github.com/Severson-Group/KnowledgeBase/blob/master/code/python/README.md#style-guidelines) and [here](https://www.python.org/dev/peps/pep-0008/)

- [x] Revise to have a contract module and a module per tool. The contract module should hold all of the abstract base classes that a tool implements. You have started doing this in `AbstractBaseClasses.py`. We would like you to rename this to follow Python style and to be specific that it applies to tools.

- [ ] Figure out best way to implement tool modules (determine the template design for a tool module and implement the MagNet tool module in that fashion). We like the idea of breaking this out into multiple submodules that the overarching tool module brings together. Is this Pythonic?

elsevers (approved):
Please go ahead and merge this and then create a new PR that implements that changes discussed at Friday's meeting.

ngadiyar93 (approved):
These changes look good to me 👍

Bharat-Ramadas:
Changes discussed on 11/30/2020 have been implement in PR #132

PR #129, opened 2020-10-23
Pass each csToken to prepareSection

### LINKED ISSUE #128: Call prepareSelection separately for each cross-section
There is a bug where when multiple cross-sections are prepared for extrusion, `prepareSection` is passed all of the cross-section tokens as a single array but called multiple times.

See here --
https://github.com/Severson-Group/eMach/blob/3a01680eb4418ec01a1a134989a37b53213d71bc/model_obj/makeSolid/MakeExtrude.m#L23-L28

We need to fix this by calling it on each cross-section individually. That is, line 24 should instead be 
```token1(i) = maker.prepareSection(csToken(i));```

ngadiyar93:
This PR fixes a bug in `MakeExtrude` where when multiple cross-sections are prepared for extrusion, prepareSection is passed all of the cross-section tokens as a single array but called multiple times.

See Issue #128 for details

### DIFF
--- a/model_obj/makeSolid/MakeExtrude.m
+++ b/model_obj/makeSolid/MakeExtrude.m
@@ -21,7 +21,7 @@ classdef MakeExtrude < MakeSolidBase
             
             %1. Prepare to extrude
             for i = 1:length(csToken)
-                token1 = maker.prepareSection(csToken);
+                token1(i) = maker.prepareSection(csToken(i));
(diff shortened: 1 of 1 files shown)


Bharat-Ramadas (approved):
Works with `exampleBreadLoad`

elsevers (approved):

PR #131, opened 2020-11-06
Add JMAG tool

### LINKED ISSUE #125: Add JMAG MATLAB tool
See PR #90. 

This should be a fully functional tool that can implements that abstract base classes `ToolBase`, `DrawerBase`, `MakerExtrudeBase`, `MakerRevolveBase` and can be used in place of the MagNet tool for drawing / making.

### LINKED ISSUE #126: Add examples that use JMAG MATLAB tool
Once #125 has been implemented, let's add examples to the eMach code base that use this tool. 

This means two things:
1. Updating the drawing/making examples in the root of `eMach/examples/`  to have a `if` switch to use JMAG instead of TikZ or MagNet

2. Adding support for a complete solve using JMAG to one or more of our FEA examples `eMach/examples/`  (i.e., 
`eMach/examples/exampleAFPM2D/` or the yet to be merged #67) to be able to use either JMAG or MagNet. Doing this may [optionally] involve creating the necessary porcelain/plumbing functions.

### LINKED ISSUE #134: Determine how to handle additional regions created when extruding parts in JMAG

### LINKED ISSUE #135: Determine appropriate workflow in JMAG for eMach
Figure out how study should be integrated into workflow.

We think that this should be handled by using the state of the JMAG tool object. Each JMAG tool object gets a single study. All components are added into that study.

### LINKED ISSUE #151: Add markdown doc explaining JMAG component creation workflow implementation and potential future JMAG version compatibility  issues
Add markdown document that explains the JMAG component creation workflow, how we are handling the extra regions that are created, and potential future JMAG version compatibility issues

In particular, discuss the assumption about how region names composed based on the number of geometry elements and how you determine which regions to delete.

ngadiyar93:
This PR adds JMAG tool to eMach.

To explore the JMAG tool, please check out this branch and run `twoOuterRotorStator.m` from the examples folder.
You should be able to see the stator of an outer rotor machine drawn in JMAG.


ngadiyar93:
@ASHAD1 The `DimMeter` is added in this PR, as well as updates to `DimMillimeter` are made here. Do you want to add those as separate PRs? Maybe we only want to limit this PR to adding the JMAG tool (and examples)?

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,348 @@
...
+        geomApp = 0; % The Geometry Editor object
+        doc = 0; % The document object in Geometry Editor
+        assembly = 0; % The assembly object in Geometry Editor
+        sketch = 0; % The sketch object in Geometry Editor
+        part = 0; % The part object in Geometry Editor
+        model = 0; % The model object in JMAG Designer
+        study = 0; % The study object in JMAG Designer
+        view;  % The view object in JMAG Designer
+        defaultLength = 'meters'; % Default length unit is m
+        defaultAngle = 'degrees'; % Default angle unit is degrees
```
elsevers: Let's make `defaultLength` and `defaultAngle` be strings that correspond to class names of dimension objects, so that `feval` can be used to do the unit conversion

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,348 @@
...
+        
+        
+        function setDefaultLengthUnit(obj, userUnit)
+            %SETDEFAULTLENGTHUNIT Set the default unit for length.
+            %   setDefaultLengthUnit(userUnit)
+            %       Sets the units for length. 
+            %   userUnit can be set to meters
+  
+            if strcmp(userUnit, 'meters')
+                obj.model.SetUnitCollection('SI_units')
```
elsevers: Need to update `self.defaultLength` to `DimMeter`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,348 @@
...
+        end
+        
+        function setDefaultAngleUnit(obj, userUnit)
+            %SETDEFAULTANGLEUNIT Set the default unit for angle.
+            %   setDefaultAngleUnit(userUnit)
+            %       Sets the units for angle. 
+            %   userUnit can be set to degrees
+  
+            if strcmp(userUnit, 'degrees')
+                obj.model.SetUnitCollection('SI_units')
```
elsevers: Need to update `self.defaultAngle` to `DimDegree`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,348 @@
...
+            else
+                error('unsupported angle unit')
+            end
+        end
+        
+        
+        function convertedLength = convertLengthUnit(obj, length, userUnit)
```
elsevers: Eliminate this function and instead just use `feval` combined with `defaultLength` inside the functions that need to convert units

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,348 @@
...
+  
+            if strcmp(userUnit, 'meters')
+                convertedLength = double(DimMeter(length));
+            else
+                error('unsupported length unit')
+            end
+        end
+        
+        
+        function convertedAngle = convertAngleUnit(obj, angle, userUnit)
```
elsevers: Eliminate this function and instead just use `feval` combined with `defaultAngle` inside the functions that need to convert units

ASHAD1:
@ngadiyar93 I agree with your earlier comment. So this PR doesn't include changes in the dimensions class. The `DimMeter` class was added in another PR to develop branch, and I have pulled the latest develop changes in this branch.

examples/twoOuterRotorStator.m
```diff
@@ -0,0 +1,110 @@
...
+        'dim_r_sb', DimMillimeter(0), ...
+        'dim_Q', 8, ...
+        'location', Location2D( ...
+            'anchor_xy', DimMillimeter([0,0]), ...
+            'theta', DimDegree(0).toRadians() ...
+        ) ...
+);
+
+stator2 = CrossSectOuterRotorStator( ...
+        'name', 'stator1', ...
```
Bharat-Ramadas: Please revise name to 'stator2'

examples/xfemmExampleArc.m
```diff
@@ -3,7 +3,8 @@
```
Bharat-Ramadas: Looks like this is an xFemm example file. Do you want to include this in the PR?

examples/xfemmExampleNotchedRectangle.m
```diff
@@ -3,7 +3,8 @@
```
Bharat-Ramadas: Looks like this is an xFemm example file. Do you want to include this in the PR?

examples/twoOuterRotorStator.m
ngadiyar93: @Bharat-Ramadas Done :-)

examples/xfemmExampleNotchedRectangle.m
ngadiyar93: Removed this

ngadiyar93:
@aimthiazz @Bharat-Ramadas I have worked to address your feedback from the review we had earlier this week.

aimthiazz (dismissed):
Checked it in my system.

Bharat-Ramadas (dismissed):
Looks good to me.

examples/outerRotorStatorExample.m
```diff
@@ -33,7 +34,7 @@
 comp1 = Component( ...
         'name', 'comp1', ...
         'crossSections', cs, ...
-        'material', MaterialGeneric('name', 'M19: USS Transformer 72 -- 29 Gage'), ...
+        'material', MaterialGeneric('name', '10JNEX900'), ...
```
elsevers: This looks like it will break MagNet. We can't merge the code if it will break another tool.

Need to fix the examples so that MagNet still works. *Please verify.*

elsevers (changes requested):
See detailed requests in the comments.

tools/jmag/JMAG.m
ASHAD1: Updated in the latest commit.

tools/jmag/JMAG.m
ASHAD1: Updated in the latest commit.

tools/jmag/JMAG.m
ASHAD1: Updated in the latest commit.

tools/jmag/JMAG.m
ASHAD1: Updated in the latest commit.

tools/jmag/JMAG.m
ASHAD1: Updated.

ASHAD1:
@elsevers I think this ready for your review.

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,324 @@
...
+            %   center - x,y coordinate of center point of rotation
+            %   axis   - x,y coordinate on the axis of ration (negative reverses
+            %             direction) (0, -1) to rotate clockwise about the y axis
+            %   angle  - Angle of rotation (dimAngular) 
+        end
+        
+        
+        function extrudeSketch = extrude(obj, name, material, depth, csToken)
+            ref1 = obj.sketch;
+            obj.part.CreateExtrudeSolid(ref1,double(DimMeter(depth)))
```
Bharat-Ramadas: Shouldn't `depth` be a DimLinear object? I think you should use `feval` here as well.

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,324 @@
...
+            obj.doc.GetSelection().Clear();
+            for i = 1:length(csToken.token)
+                obj.doc.GetSelection().Add(obj.sketch.GetItem(csToken.token(i).drawToken.GetName()));
+            end
+            id = obj.sketch.NumItems();
+            obj.sketch.CreateRegions();
+            id2 = obj.sketch.NumItems();
+            visItem = 1; % Set 1 to select only the visible (top layer) item
+            itemType = 64; % Set 64 for region.
+            obj.geomApp.View.SelectAtCoordinateDlg(double(DimMeter(csToken.innerCoord(1))), ...
```
Bharat-Ramadas: Should this be `feval` instead of `DimMeter(csToken.innerCoord(1))`?

Bharat-Ramadas (commented):
Nice job @ASHAD1. I think the PR is almost done, but I have a few comments. Please see below for further details

tools/jmag/JMAG.m
ngadiyar93: Thanks @Bharat-Ramadas I have updated this.

tools/jmag/JMAG.m
ngadiyar93: Thanks @Bharat-Ramadas I have updated this.

Bharat-Ramadas (dismissed):
Looks good

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,329 @@
...
+        
+        function new = revolve(obj, name, material, center, axis, angle)
+            %REVOLVE Revolve a cross-section along an arc    
+            %new = revolve(obj, name, material, center, axis, angle)
+            %   name   - name of the newly extruded component
+            %   center - x,y coordinate of center point of rotation
+            %   axis   - x,y coordinate on the axis of ration (negative reverses
+            %             direction) (0, -1) to rotate clockwise about the y axis
+            %   angle  - Angle of rotation (dimAngular) 
+        end
```
elsevers: Contents of the revolve function are missing

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,329 @@
...
+        
+        
+        function extrudeSketch = extrude(obj, name, material, depth, csToken)
```
elsevers: Please add a couple more comments in this function explaining the process to extrude

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,329 @@
...
+            obj.setDefaultLengthUnit(obj.defaultLength)
+            obj.setDefaultAngleUnit(obj.defaultAngle)
+            % Add material
+            obj.study.SetMaterialByName(name, material)
+            obj.app.Save()
+            extrudeSketch = 0;
+        end
+        
+        
+        function sketch = prepareSection(obj, csToken)
```
elsevers: Please add a couple more comments explaining the what is taking place in this function

elsevers (changes requested):

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+            %REVOLVE Revolve a cross-section along an arc    
+            %new = revolve(obj, name, material, center, axis, angle)
+            %   name  - name of the newly revolved component
+            %   center - x,y coordinate of center point of rotation
+            %   axis   - x,y coordinate on the axis of ration (negative reverses
+            %             direction) (0, -1) to rotate clockwise about the y axis
+            %   angle  - Angle of rotation (dimAngular) 
+            
+            % Convert to default units
+            center = double(feval(obj.defaultLength, center));
```
Bharat-Ramadas: Is there a reason for converting the DimLinear objects to double?

Bharat-Ramadas (commented):
Please see my comment below

tools/jmag/JMAG.m
ASHAD1: The arguments that are passed to the JMAG functions need to be data type double.

elsevers: This does not make sense. All dimensions are type `double`. We should not be converting them to doubles from their respective classes.

elsevers (changes requested):
`twoOuterRotorStator.m` example does not work. I immediately get a MATLAB error, see below. It looks like you have hard coded this to only work for JMAG 18.1. This code needs to work with any version of JMAG. **Please fix**

image

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+        sketch = 0; % The sketch object in Geometry Editor
+        part = 0; % The part object in Geometry Editor
+        model = 0; % The model object in JMAG Designer
+        study = 0; % The study object in JMAG Designer
+        view;  % The view object in JMAG Designer
+        defaultLength = 'DimMeter'; % Default length unit is m
+        defaultAngle = 'DimDegree'; % Default angle unit is degrees
+        visible = true; % Visibility
+        workDir = pwd;
+        sketchList;
```
elsevers: Please review your properties: make sure that all of these are needed and update the property names to be more consistent with each other and meaningful.

In particular, please rename `geomApp` (perhaps `geometryEditor`?). Why do we have both `app` and `jd`? This looks like a duplicate reference to the same object.

Why are you setting these to 0 when they are not used? Please revise to be consistent with how the `MagNet` tool is implemented (they are simply not initialized)

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+            obj.validateProps();
+            obj.jd = actxserver('designer.Application.191');
+            if obj.visible
+                obj.jd.Show();
+            else
+                obj.jd.Hide();
+            end
+        end
+        
+        function obj = open(obj, varargin)
```
elsevers: Please update this to implement the function signature in the newly revised `ToolBase`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+        end
+        
+        function obj = delete(obj)
+            % DELETE Closes the JMAG application
+            % delete()
+            
+            obj.jd.Quit();
+        end
+        
+        function save(obj, path, fileName)
```
elsevers: Please revise to use the new function signature in the updated `ToolBase`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+            % SAVE Saves the MagNet document in the specified path
+            
+            obj.jd.SaveAs(strcat(path, '\',fileName,'.jproj'));
+        end
```
elsevers: Please add a `saveAs` function

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+            %   drawLine([start_x, _y], [end_x, _y]) draws a line
+
+            if isnumeric(obj.sketch)
+                obj.sketch = obj.getSketch(0);
+                obj.sketch.OpenSketch();
+            end
+            
+            % Convert to default units
+            startxy = double(feval(obj.defaultLength, startxy));
+            endxy = double(feval(obj.defaultLength, endxy));
```
elsevers: These should not be converted to `double`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+        
+        function [tokenDraw] = drawArc(obj, centerxy, startxy, endxy)
+            %DRAWARC Draw an arc in the current JMAG document.
+            %   drawarc(mn, [center_x,_y], [start_x, _y], [end_x, _y])
+            %       draws an arc
+            
+            if isnumeric(obj.sketch)
+                obj.sketch = obj.getSketch(0);
+                obj.sketch.OpenSketch();
+            end
```
elsevers: Why do we have to do this check every time `drawArc` or `drawLine` are called? Shouldn't the `sketch` property get initialized in the `open` function and then never have to be updated?

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+                obj.study = obj.model.GetStudy(obj.projName);
+            end
+            
+            % Set to default units
+            obj.setDefaultLengthUnit(obj.defaultLength)
+            obj.setDefaultAngleUnit(obj.defaultAngle)
+            
+            % Add material
+            obj.study.SetMaterialByName(name, material)
+            obj.app.Save()
```
elsevers: `obj.app.Save()`

Why are you saving after you extrude/revolve a part? Wouldn't it be better for the user to call the save function for the tool? (What if a user wants to extrude something but not save the results?)

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,399 @@
...
+            
+            for idx = 1:id2-id
+                if ~strcmp(regionList{idx}, regionName)
+                    obj.doc.GetSelection().Clear();
+                    obj.doc.GetSelection().Add(obj.sketch.GetItem(regionList{idx}));
+                    obj.doc.GetSelection().Delete();
+                end
+            end          
+            obj.sketch.CloseSketch();
+            sketch = 1;
```
elsevers: `sketch = 1;`

I don't understand what is going on with your `sketch` property and why it is getting updated like this.

tools/jmag/README.md
```diff
@@ -0,0 +1,46 @@
...
+This document briefly explains how the component creation workflow is handled in JMAG.
+
+### Background
+The geometry creation workflow in eMach is as follows:
+
+1. Draw a cross section geometry.
+2. Extrude the cross section - requires a coordinate that is inside the region to be extruded. 
+3. Assign materials.
+
+Although this workflow is consistent with the workflow in Infolytica MagNet, the workflow in JMAG has some additional steps to be followed. The JMAG workflow is as follows:
```
elsevers: Don't be talking about MagNet here. Just talk about eMach and JMAG.

tools/jmag/README.md
```diff
@@ -0,0 +1,46 @@
+## JMAG Tool
+
+This document briefly explains how the component creation workflow is handled in JMAG.
+
+### Background
+The geometry creation workflow in eMach is as follows:
+
+1. Draw a cross section geometry.
+2. Extrude the cross section - requires a coordinate that is inside the region to be extruded.
```
elsevers: This item 2 is written too casually. Revise your language to be more professional.

tools/jmag/README.md
```diff
@@ -0,0 +1,46 @@
...
+Although this workflow is consistent with the workflow in Infolytica MagNet, the workflow in JMAG has some additional steps to be followed. The JMAG workflow is as follows:
+1. Draw a cross section geometry.
+2. Create regions.
+3. If more than one region is created in Step 2, delete unecessary regions.
+4. Extrude the region.
+5. Create a study.
+6. Add the geometry to the study.
+7. Assign materials.
+
+In the JMAG workflow, steps 2. and 3. need to be suitably handled. This is because, the JMAG function to create regions simply converts any closed geometry (bounded by lines and arcs) into a JMAG geometry region.
```
elsevers: `steps 2. and 3.` 
Don't put periods after these numbers

`This is because, the...` - language and grammar revision are needed

`JMAG geometry region` -- language issue here. What is noun you mean? Does simply `region` (no `JMAG geometry`)?

tools/jmag/README.md
```diff
@@ -0,0 +1,46 @@
...
+1. Draw a cross section geometry.
+2. Create regions.
+3. If more than one region is created in Step 2, delete unecessary regions.
+4. Extrude the region.
+5. Create a study.
+6. Add the geometry to the study.
+7. Assign materials.
+
+In the JMAG workflow, steps 2. and 3. need to be suitably handled. This is because, the JMAG function to create regions simply converts any closed geometry (bounded by lines and arcs) into a JMAG geometry region. 
+For example, if a stator cross-section of a radial flux machine is drawn, JMAG creates two regions - one the actual stator region comprising the teeth, yoke etc, and the other region comprising the stator bore and slots.
```
elsevers: Language issues/revision needed

tools/jmag/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## The solution
+If working with the JMAG's gui, one can select the region that is not required and delete it manually. However, this is hard to do programmatically because of the following reasons:
```
elsevers: `gui` should be in all capital letters.

tools/jmag/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+1. The `CreateRegions()` function in JMAG takes no arguments - i.e, there is no way to specify where to create JMAG geometry regions. All closed geometries are simply converted to JMAG geometry regions.
```
elsevers: This does not belong here: `- i.e.`. Please revise your language to be more professional.

elsevers (changes requested):
Please resolve all merge conflicts. Then, merge the latest `develop` branch into this branch and make the JMAG tool compliant with the latest version of the `ToolBase` class.

Go through my detailed comments below. 

Finally, go function by function and compare your JMAG tool to the latest version of the MagNet tool. Think about what differences there should be between the MATLAB implementation of the two tools and what should be the same.

tools/jmag/JMAG.m
ASHAD1: I am not sure about that unless it's converted to double JMAG was throwing an error.

ngadiyar93: @Bharat-Ramadas and @elsevers I agree. However, when I checked, I too found that @ASHAD1 is right -- unless converted to double, JMAG throws an error. 

The following test script can be used to quickly check this:

```matlab
%%  Initialize JMAG
toolJd = JMAG();
toolJd.setVisibility(true);
toolJd.open();
toolJd.sketch = toolJd.getSketch(0);
toolJd.sketch.OpenSketch();

%%  Draw the first vertex with coords in double
a = double(DimMeter(1));
b = double(DimMeter(2));
toolJd.sketch.CreateVertex(a, b);

%%  Draw a second vertex in DimMeters
c = DimMeter(3);
d = DimMeter(4);
toolJd.sketch.CreateVertex(c, d);

```

It runs fine until the last line, but throws the following error on the last line:
image

I am guessing this could be because we are using JMAG functions directly, and JMAG has no info on the `DimMeter` data type. This is unlike MagNet where we do an invoke and pass in a string: 
image

tools/jmag/JMAG.m
ASHAD1: I have implemented this.

tools/jmag/JMAG.m
elsevers: Unless anyone else has ideas on how to debug further, it is okay to proceed by converting these to `double`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,394 @@
...
+            if isempty(obj.fileName)
+                error('Unable to save file. Use the saveAs( ) function');
+            else
+                obj.jd.SaveAs(obj.fileName);
+            end        
+        end
+        
+        
+        function obj = close(obj)
+            % CLOSE Closes the JMAG document
```
ASHAD1: I have not implemented anything in the `close` function because there is nothing equivalent to closing the document in JMAG. JMAG supports just one project per instance. I have illustrated this using the following example.

Let's say if have a project opened in JMAG and if I try to open a new project as shown in the following screenshot. 
Screenshot (26)

When I create a new project, it automatically closes the previous project and opened a new project.

I am running the script recording, using this we can see via scripts what JMAG is doing. In the following screenshot it can be seen that the script recorded only creating a new project, it didn't record any closing command.
image

So I think JMAG currently doesn't support command for closing a project. @elsevers Please let me know your suggestions on this.

tools/jmag/JMAG.m
ASHAD1: For every new component, a new sketch (and part) is created. I have updated this part of the code. Please review the updated code.

tools/jmag/JMAG.m
Bharat-Ramadas: Hello @ASHAD1. I was going through the JMAG Script help and found a `CloseInstance( )` function there. Maybe this can be used here? 

image

Also please find below the definition for the Quit( ) function we normally employ when closing JMAG.
image

And finally this is the entire list of Public Member Functions available to the `InstanceManager` class in JMAG.
image

Bharat-Ramadas (commented):
Hello @ASHAD1, @ngadiyar93. Nice work! I have just one comment on the close( ) function.

tools/jmag/JMAG.m
ASHAD1: Interesting find @Bharat-Ramadas. Now I am using this `InstanceManager()` for opening and quitting the JMAG application now. Previously when I was closing JMAG application using `delete(toolJd)`, all the JMAG instances previously opened were closing. Now if I call `delete(toolJd)` (similar to Magnet implementation) it will now only quit the JMAG application instance opened from Emach. 

However, this will not help the `close()` implementation. As per Magnet implementation `close()` should close the document, not the JMAG application instance itself. The Jmag application instance is closed when `delete()` is called. Therefore I have not implemented `close()` in JMAG. @elsevers please let me know if this fine for now.

Bharat-Ramadas (dismissed):
Looks good!

tools/jmag/JMAG.m
elsevers: A user expects that if they call `close` that future function calls will not edit the document they previously had open. If JMAG does not have a notion of document close feature, the proper functionality of `close` should be to open a new document. That is, `close()` should call `obj.open()` (open an untitled file).

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+            end
+        end
+        
+        
+        function obj = open(obj, fileName)
+            %OPEN Open JMAG Designer or a specific file.
+            %   open(fileName) opens an JMAG document. 
+            %   fileName is a string that specifies the complete path to the file.
+            if ~exist('fileName', 'var')
+                error('Please specify a filename');
```
elsevers: You currently have this set up so that `filename` is required. However, we'd like to be able leave `filename` as optional so that a user can create an unsaved project. 

In JMAG Designer, if I do `file`->`new` I see this scripting result:
```
Set app = designer
Call app.NewProject("Untitled")
```

If I then click save, JMAG Designer prompts me for a filename to save it as. So it does appear that JMAG allows you to create a new, unsaved project. We should support this as well.

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+        sketch; % The sketch object in Geometry Editor
+        part; % The part object in Geometry Editor
+        model; % The model object in JMAG Designer
+        study; % The study object in JMAG Designer
+        view;  % The view object in JMAG Designer
+        defaultLength = 'DimMeter'; % Default length unit is m
+        defaultAngle = 'DimDegree'; % Default angle unit is degrees
+        visible = true; % Visibility
+        fileName;
+    end
```
elsevers: Why are we saving `jdInstance`? It looks like this is never actually used. Let's get rid of it.

`geometryEditor` also looks like something we do not need to be saving. The one time it is used is the context of `obj.geometryEditor.view.SelectAtCoordinateDlg()` -- couldn't you just use `obj.view.SelectAtCoordinateDlg()`?

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+            if strcmp(userUnit, 'DimDegree')
+                obj.defaultAngle = userUnit;
+                obj.model.SetUnitCollection('SI_units')
+            else
+                error('unsupported angle unit')
+            end
+        end
+        
+        
+        function setVisibility(obj, visibile)
```
elsevers: You have misspelled visible here

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+            obj = obj.createProps(nargin,varargin);            
+            obj.validateProps();
+            % Obtain instance of JMAG designer application
+            obj.jdInstance = actxserver('designerstarter.InstanceManager');
+            obj.jd = obj.jdInstance.GetInstance(); % Creates a new instance and returns the handle
+            if obj.visible
+                obj.jd.Show();
+            else
+                obj.jd.Hide();
+            end
```
elsevers: Keep your code DRY. Instead of
```matlab
            if obj.visible
                obj.jd.Show();
            else
                obj.jd.Hide();
            end
```

Call `obj.setVisibility(obj.visible);`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+            %   open(fileName) opens an JMAG document. 
+            %   fileName is a string that specifies the complete path to the file.
+            if ~exist('fileName', 'var')
+                error('Please specify a filename');
+            elseif ~exist(fileName, 'file')
+                obj.jd.NewProject('proj');
+                obj.saveAs(fileName);
+            else
+                obj.jd.Load(fileName);
+            end
```
elsevers: You forgot to update `obj.fileName` when you call `obj.jd.Load()`

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+            obj.view = obj.jd.View();
+            obj.geometryEditor = obj.jd.CreateGeometryEditor(true);
+            obj.doc = obj.geometryEditor.NewDocument();
+            obj.assembly = obj.doc.GetAssembly();
+        end
```
elsevers: I am not familiar enough with JMAG to be certain about this, but it seems like there probably should be a function call to `setDefaultLengthUnit` and `setDefaultAngleUnit` when you open a new document.

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
...
+            obj.doc.SaveModel(true)
+            modelName = strcat(name,'_model');
+            studyName = strcat(name,'_study');
+            
+            % Create a study
+            obj.study = obj.createStudy(studyName, modelName);
+            
+            % Set to default units
+            obj.setDefaultLengthUnit(obj.defaultLength)
+            obj.setDefaultAngleUnit(obj.defaultAngle)
```
elsevers: Is it really necessary to call these functions here?

```matlab
            obj.setDefaultLengthUnit(obj.defaultLength)
            obj.setDefaultAngleUnit(obj.defaultAngle)
```

Or can they just be called upon opening a file?

tools/jmag/JMAG.m
```diff
@@ -0,0 +1,367 @@
+classdef JMAG < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
```
elsevers: Since there are multiple JMAG tools, we had better be explicit that this tool is for JMAG-Designer. Please rename this class as `JMAG_Designer`.

tools/jmag/README.md
```diff
@@ -0,0 +1,44 @@
...
+The workflow for geometry creation in JMAG is listed below:
+1. Draw a cross section geometry.
+2. Create regions.
+3. If more than one region is created in Step 2, delete unnecessary regions.
+4. Extrude the region.
+5. Create a study.
+6. Add the geometry to the study.
+7. Assign materials.
+
+Although the geometry creation workflow in JMAG is quite consistent with the eMach workflow, one of the important differences is the creation of regions.
```
elsevers: Let's cut this first sentence. It does not make sense to try to parse out whether there is a difference between the JMAG and eMach workflow.

tools/jmag/README.md
```diff
@@ -0,0 +1,44 @@
...
+3. If more than one region is created in Step 2, delete unnecessary regions.
+4. Extrude the region.
+5. Create a study.
+6. Add the geometry to the study.
+7. Assign materials.
+
+Although the geometry creation workflow in JMAG is quite consistent with the eMach workflow, one of the important differences is the creation of regions. 
+The JMAG function to create regions simply converts any closed geometry (bounded by lines and arcs) into a region. For example, if the stator cross-section of a radial flux machine is drawn, JMAG creates two regions: i) `region 1`: a region consisting the stator teeth and yoke, and ii) `region 2`: a region with the stator bore and slots. If this geometry is extruded, it will result in a solid cylinder. So, the unnecessary region (`region 2`) must be identified and deleted before extrusion.  
+
+## The solution
```
elsevers: Rename this heading to `## Implementation in eMach`

tools/jmag/README.md
```diff
@@ -0,0 +1,44 @@
...
+If working with the JMAG's GUI, one can select the unnecessary region and delete it manually. However, this is hard to do programmatically because of the following reasons:
+
+1. The `CreateRegions()` function in JMAG takes no arguments. So, there is no way to specify where to create JMAG geometry regions. All closed geometries are simply converted to JMAG geometry regions.
+2. The `CreateRegions()` function does not return the name or any other information about the regions created. 
+
+The eMach workflow provides a coordinate `innerCoord` inside every cross section to be extruded. This `innerCoord` will be inside a geometry region that must be retained and extruded in JMAG. With this information, the following workaround was adopted to address the issue of multiple geometry regions:
```
elsevers: Instead of _workaround was adopted to address the issue of multiple geometry regions:_

say _approach is used to create only the desired geometry region:_

tools/jmag/README.md
```diff
@@ -0,0 +1,44 @@
...
+3. The number of geometry items after region creation is obtained by calling `NumItems()`.
+4. The difference between the number of geometry items in Step 3 and Step 1 gives the total number of regions created in Step 2.
+5. JMAG names regions in the format `region, region.2, region.3,......region.n` if there are `n` regions. So, knowing the number of regions created, an array of region names is populated.
+6. The `innerCoord` is used with the `SelectAtCoordinateDlg()` function to select the desired region to be kept.
+7. The `GetSelection()` function is used to get the name of the selected region which is to be retained.
+8. The region name obtained in step 7 is compared with the array of region names populated in Step 5. All unnecessary regions created by JMAG are deleted.
+
+The desired region is left behind and can then be extruded or revolved.
+
+**Note:** The steps listed above to delete the undesired regions heavily rely on JMAG's region naming convention. If future versions of JMAG updates the way geometry regions are named, this must be updated accordingly.
```
elsevers: Remove the word _heavily_

Change _updates the way geometry regions are named_ to _update the naming of geometry regions_

Change _this must be updated accordingly_ to _the JMAG implementation will need to be revised._

elsevers (changes requested):
This is getting closer, but several issues remain that need to be resolved. See the bug I report here as well as the in-line comments for `JMAG.m`

When I create multiple objects of the JMAG class, each object is re-using the same JMAG instance. For example, if I run this code:
```matlab
toolJd1 = JMAG();
toolJd2 = JMAG();
```
I want that two JMAG windows to open up. But only one window opens. This is because both `toolJd1` and `toolJd2` are re-using the same JMAG application instance. Please fix this. We want to be able to have multiple JMAG windows open at one time.
This is not the

elsevers:
I also don't think that the `delete` function is working properly. If I clear my MATLAB workspace (`clear`), JMAG does not exit. However, if I try the same thing with MagNet, MagNet exits.

tools/jmag/JMAG.m
ASHAD1: This is implemented.

tools/jmag/JMAG.m
ASHAD1: This is implemented.

tools/jmag/JMAG.m
ASHAD1: I have removed `jdInstance` from the properties. But I was not able to do it for `geometryEditor`. It will throw an error if I just use `obj.view.SelectAtCoordinateDlg()` instead of `obj.geometryEditor.view.SelectAtCoordinateDlg()`. Let me know if this is fine.

tools/jmag/JMAG.m
ASHAD1: In JMAG, `units` are associated with the JMAG `model`. We cannot assign units until we have the `model` imported from the `Geometry Editor` to the `JMAG Designer`. Setting units in JMAG is a bit tricky, I have implemented the workflow for setting units as per the workflow we had agreed upon. Let me know if this is fine.

tools/jmag/JMAG.m
ASHAD1: In JMAG, `units` are associated with the JMAG `model`. We cannot assign units until we have the `model` imported from the `Geometry Editor` to the `JMAG Designer`. Setting units in JMAG is a bit tricky, I have implemented the workflow for setting units as per the workflow we had agreed upon. Let me know if this is fine.

tools/jmag/JMAG.m
ASHAD1: This is implemented

ASHAD1:
> I also don't think that the `delete` function is working properly. If I clear my MATLAB workspace (`clear`), JMAG does not exit. However, if I try the same thing with MagNet, MagNet exits.

I have fixed this issue in the latest commit.

tools/jmag/JMAG_Designer.m
```diff
@@ -0,0 +1,375 @@
...
+        end
+        
+        
+        function open(obj, fileName)
+            %OPEN Open JMAG Designer or a specific file.
+            %   open(fileName) opens an JMAG document. 
+            %   fileName is a string that specifies the complete path to the file.
+            if ~exist('fileName', 'var')
+                obj.jd.NewProject('untitled');
+                obj.saveAs('untitled'); % JMAG requires it to be saved before creating geometry
```
elsevers: No, we can't assume a location to save a file. This will lead to strange and unexpected behavior down the road.

After doing some digging here, I think you need to go back to requiring that user provide a filename here. You can make it be invalid to call `open` without passing in a file name (to either open or to create). 

Please update your documentation for this function to indicate this. Make sure you follow the [function documentation style for eMach.](https://github.com/severson-group/emach#in-function-documentation)

tools/jmag/JMAG_Designer.m
```diff
@@ -0,0 +1,375 @@
...
+            endxy = double(feval(obj.defaultLength, endxy));             
+            
+            line = obj.sketch.CreateLine(startxy(1),startxy(2),endxy(1),endxy(2));
+            tokenDraw = TokenDraw(line, 0);
+        end
+        
+        
+        function [tokenDraw] = drawArc(obj, centerxy, startxy, endxy)
+            %DRAWARC Draw an arc in the current JMAG document.
+            %   drawarc(mn, [center_x,_y], [start_x, _y], [end_x, _y])
```
elsevers: This function signature comment specifies `mn` as an argument. This is not correct.

tools/jmag/JMAG_Designer.m
```diff
@@ -0,0 +1,375 @@
...
+        function study = createStudy(obj, studyName, modelName)
+            % CREATESTUDY Creates a new study
+            % study = createStudy(studyName, modelName)
+            % studyName - Name of the study to be created
+            % modelName - Model where the study will be created
+            if isempty(obj.study)
+                obj.model = obj.jd.GetCurrentModel();
+                obj.model.SetName(modelName)
+                % Create study
+                study = obj.model.CreateStudy('Transient', studyName);
```
elsevers: How do you know that the study should be `transient`? What does this do?

tools/jmag/JMAG_Designer.m
```diff
@@ -0,0 +1,375 @@
...
+        view;  % The view object in JMAG Designer
+        defaultLength = 'DimMeter'; % Default length unit is m
+        defaultAngle = 'DimDegree'; % Default angle unit is degrees
+        visible = true; % Visibility
+        fileName;
+    end
+    
+
+    methods
+        function obj = JMAG_Designer(varargin)
```
elsevers: After doing some digging on how JMAG handles files, I think you need to go back to requiring that user provide a filename in the initializer. You should then call `obj.open(filename)` in your initializer.

Please update your documentation for this function to indicate this. Make sure you follow the [function documentation style for eMach.](https://github.com/severson-group/emach#in-function-documentation)

tools/jmag/JMAG_Designer.m
```diff
@@ -0,0 +1,375 @@
...
+            % SAVE Saves the JMAG document in the specified path
+            if isempty(obj.fileName)
+                error('Unable to save file. Use the saveAs( ) function');
+            else
+                obj.jd.SaveAs(obj.fileName);
+            end        
+        end
+        
+        
+        function close(obj)
```
elsevers, line 110: Yikes. What to do here since our instance of this object always needs to have a valid file? I think that for now, `close` can just call `delete`.

elsevers (changes requested):
Please update your function comments to follow the [eMach function comment guidelines](https://github.com/severson-group/emach#in-function-documentation)

--------------

I tried modifying the `exampleLinearMotor.m` file to use this tool and ran into issues.

Here's what I did:

1. Run `exampleLinearMotor.m` -- this will draw the linear machine in MagNet
2. Run these commands:
```matlab
fileName = strcat(path,'\test3.jproj');
toolJd = JMAG_Designer();
toolJd.setVisibility(true);
toolJd.open(fileName);
statorIronComp.make(toolJd, toolJd);
moverIronComp.make(toolJd, toolJd);
magnet1Comp.make(toolJd, toolJd);
magnet2Comp.make(toolJd, toolJd);
magnet3Comp.make(toolJd, toolJd);
```

The end result seems to have two problems:
1. It leaves a JMAG-Designer Geometry editor window open. Is this okay?
2. It has created separate solid models for every single component that was extruded -- see screenshot below:

image

tools/jmag/JMAG.m
elsevers: Yes, this is okay.

tools/jmag/JMAG_Designer.m
ASHAD1: The JMAG `study` needs to be created for assigning materials, and we need to specify the type of study (`Transient`, `Static`) when we are creating the study and this `studyType` cannot be changed later.
To solve this issue, I have added the ability for the user to define the `studyType` when calling the `JMAG_Designer` initializer using the key-value pair argument (varargin). Please refer to the `JMAG_Designer` initializer method for more information.

ASHAD1:
I have updated the code that will resolve the issues (raised by @elsevers).

- Geometry Editor application will now be closed as the `close()` function is now calling `delete()` function which will close both JMAG Designer and Geometry Editor applications.
- The current implementation solves the issue of multiple model creation.
- I have updated the function comments as per the guidelines
- The JMAG `study` needs to be created for assigning materials, and we need to specify the type of study (`Transient`, `Static`) when we are creating the study and this cannot be changed later.
To solve this issue, I have added the ability for the user to define the `studyType` when calling the `JMAG_Designer` initializer using the key-value pair argument (varargin). Please refer to the `JMAG_Designer` initializer method for more information.

tools/jmag/JMAG_Designer.m
```diff
@@ -0,0 +1,465 @@
...
+           
+            lenVarargin = length(varargin);
+            obj = obj.createProps(lenVarargin,varargin);            
+            obj.validateProps();
+            
+            for k = 1:lenVarargin
+                if strcmpi(varargin{k},'studyType')
+                    obj.studyType = varargin{k+1}; 
+                end
+            end
```
Bharat-Ramadas: Do you need this for loop for assigning a value to the `studyType `property? It seems to me that `obj.createProps` will take care of this.

Bharat-Ramadas (commented):
Looks good. I just have one comment on the initializer. Please find it below

Bharat-Ramadas (approved):
Looks good

Bharat-Ramadas (dismissed):
Looks good

elsevers (changes requested):
This is looking pretty good. However, I seem to have found a problem.

If I run `exampleBreadLoaf.m` , a JMAG project file is created `test_breadLoaf.jproj` and JMAG closes. If I then try to open this JMAG file by running this command `toolJD = JMAG_Designer('test_breadLoaf.jproj')`, I get this prompt:

image

Seems like a bug...

ngadiyar93:
> This is looking pretty good. However, I seem to have found a problem.
> 
> If I run `exampleBreadLoaf.m` , a JMAG project file is created `test_breadLoaf.jproj` and JMAG closes. If I then try to open this JMAG file by running this command `toolJD = JMAG_Designer('test_breadLoaf.jproj')`, I get this prompt:
> 
> image
> 
> Seems like a bug...

@elsevers I faced this error when i was running `exampleBreadLoaf.m` with my MATLAB path being on a Google Drive file stream location (This is where i had cloned eMach). 

However, it runs fine if I run it with a MATLAB path that is on my disk. I also checked if the issue is due to spaces in any of the location names in the file path, but that doesn't seem to be an issue.

So to verify if this is indeed the issue, could you change line 116 in `exampleBreadLoaf.m`  to something like `path = C:\Users\username\Documents` or any other path on your disk and run it to see if JMAG still shows the same error in saving the file?

elsevers:
> 
> 
> > This is looking pretty good. However, I seem to have found a problem.
> > If I run `exampleBreadLoaf.m` , a JMAG project file is created `test_breadLoaf.jproj` and JMAG closes. If I then try to open this JMAG file by running this command `toolJD = JMAG_Designer('test_breadLoaf.jproj')`, I get this prompt:
> > image
> > Seems like a bug...
> 
> @elsevers I faced this error when i was running `exampleBreadLoaf.m` with my MATLAB path being on a Google Drive file stream location (This is where i had cloned eMach).
> 
> However, it runs fine if I run it with a MATLAB path that is on my disk. I also checked if the issue is due to spaces in any of the location names in the file path, but that doesn't seem to be an issue.
> 
> So to verify if this is indeed the issue, could you change line 116 in `exampleBreadLoaf.m` to something like `path = C:\Users\username\Documents` or any other path on your disk and run it to see if JMAG still shows the same error in saving the file?

This error is because I am not passing in the full path of the file location. However, this exposes a problem in your `open` function. If I give it a name of a file that does not exist, it should be trying to create that file. Instead, it is calling `obj.jd.Load()` on this non-existent file

ngadiyar93:
> > > This is looking pretty good. However, I seem to have found a problem.
> > > If I run `exampleBreadLoaf.m` , a JMAG project file is created `test_breadLoaf.jproj` and JMAG closes. If I then try to open this JMAG file by running this command `toolJD = JMAG_Designer('test_breadLoaf.jproj')`, I get this prompt:
> > > image
> > > Seems like a bug...
> > 
> > 
> > @elsevers I faced this error when i was running `exampleBreadLoaf.m` with my MATLAB path being on a Google Drive file stream location (This is where i had cloned eMach).
> > However, it runs fine if I run it with a MATLAB path that is on my disk. I also checked if the issue is due to spaces in any of the location names in the file path, but that doesn't seem to be an issue.
> > So to verify if this is indeed the issue, could you change line 116 in `exampleBreadLoaf.m` to something like `path = C:\Users\username\Documents` or any other path on your disk and run it to see if JMAG still shows the same error in saving the file?
> 
> This error is because I am not passing in the full path of the file location. However, this exposes a problem in your `open` function. If I give it a name of a file that does not exist, it should be trying to create that file. Instead, it is calling `obj.jd.Load()` on this non-existent file

@elsevers If you specify the full file path as the fileName (which is what the JMAG tool looks for), even if such a file is non existent, JMAG creates it. 
image
The problem is when one only specifies a file name.

For example:
`toolJd = JMAG_Designer('C:\Users\username\Documents\JMAG test\test.jproj');` works and will create a `test.jproj` if it doesn't already exist.
However, `toolJd = JMAG_Designer('test.jproj');` does not work.

I am not sure this is a bug as we are specifying that the user has to input the full file path?

elsevers:
It looks like your issue can be resolved by adding this at the start of the function:

```matlab
fileName = which(fileName);
```

Be careful about your error checking for `~exist('fileName', 'var')`

elsevers:
I see you have added my suggestion to your constructor. However, it belongs in your `open` function, not the constructor.

ASHAD1:
@elsevers I have implemented the changes you have requested in the latest commits.

elsevers:
> 
> 
> @elsevers I have implemented the changes you have requested in the latest commits.

@ASHAD1, your changes break the case of creating a new project. Try running `exampleBreadloaf.m`. It does not work.

ASHAD1:
@elsevers Thanks for pointing this out. I have fixed this issue. I have verified it by running the example files.

Also in the latest commits, for the case where the file name without the full file path is used (ex. `JMAG_Designer('test2.jproj')`), and if this file (`test2.jproj`) is not present in the current folder, then a new file will be created in the current folder. Please let me know if you want a different approach.

elsevers (changes requested):
Getting closer. Now the functionality just described works. However, now the tool is unable to open a project that is not on the MATLAB path. 

To re-create this issue, perform the following steps:

1. Setup your MATLAB path to include all the eMach project folders except `examples`. When your MATLAB working directory is set to the eMach folder, if your MATLAB path is configured correctly, the `examples` directory should be gray while the others are solid, like this screen shot: 
image

2. Set your working directory to be the `eMach\examples` folder and run `exampleBreadloaf.m`. This will create `test_breadLoaf.jproj`
3. Now set your working directory to be `eMach` (no longer in the examples folder) and try to open `test_breadLoaf.jproj` by running this command (modify your path appropriately)
```matlab
toolJD = JMAG_Designer('D:\WorkingCopies\eMach\examples\test_breadLoaf.jproj')
```
It will fail (but it should work!). You should also be able to open the file by providing a path relative to your working directory. I.e., this command should work as well (but does not)

```matlab
toolJD = JMAG_Designer('examples\test_breadLoaf.jproj')
```

Please fix the tool so that both work.

ASHAD1:
@elsevers I have fixed the issue in the latest commit.

elsevers (approved):
Looks good @ASHAD1! 

Momentous occasion: this PR is **approved**. Thanks for all of your efforts on this.

Bharat-Ramadas (approved):
🥂

PR #122, opened 2020-10-02
Add getParameter, setParameter, makeComponentInALine, makeMotionComponent functions

Bharat-Ramadas:
# Description

This PR is to merge the python implementation of certain MagNet support functions already implemented in Matlab. 

# Supported Functions

## `getParameter`
This function allows the user to get the value and the type of a parameter of a MAGNET object. The syntax is `getParameter(MN, path, parameter)` , where `MN` is the MAGNET object,  `path` is the path of the object who's parameter the user wants to read,  `parameter` is the parameter which is to be read.
## `setParameter` 
This function allows the user set the parameter of a MAGNET object with the value that the user inputs. It does not return any value. The syntax is `setParameter(Doc, opath, param, value, MNConsts)`, where `Doc` is the MAGNET Doc object, `opath` is the path of the object who's parameter is to be modified, `param` is the parameter which is to be modified, `value` is the value with which the parameter is to be updated and `MNConsts` is the MAGNET Constants object.
## `makeComponentInALine`
This function allows the user to extrude a selected section, assign it a name and a material. The syntax is `makeComponentInALine(MN, MNConsts, sweepDist, sectionName, material)`, where `MN` is the MAGNET object, `MNConsts` is the MAGNET Constants object, `sweepDist` is the height to which the user wishes to extrude the section, `sectionName` is the name the user wishes to assign to the components and `material` is the material the user wishes to assign the component. The function returns a value of 0 if the extrusion was successful and 1 otherwise.
## `makeMotionComponent`
This function allows the user to set components as moving in MAGNET . The syntax is `makeMotionComponent(Doc,ArrayOfValues)`, where Doc is the MAGNET Doc object, `ArrayOfValues` is the list of components which are to be set in motion. The function returns the path of the created motion object.
##`makeSimpleCoil`
This function allows the user to make a simple coil in MAGNET . The syntax is `makeSimpleCoil(Doc, ProblemID, ArrayOfValues)`, where `Doc` is the MAGNET Doc object, `ProblemID` is the ProblemID assigned to the coil, `ArrayOfValues` is the list of components which are to be made into a coil. The function returns the path of the created coil object.

# Problems Encountered 

- In order for Python to send commands to MAGNET , we need to import the pywin32 to call ActiveX compliant programs as COM automation servers. pywin32 module comes installed when you install Anaconda3, however for Python to Import the module, the path of the module has to be defined in the system`Environment variables` for Windows (or the file we create has to be in the same directory as that of the file we are importing). 
[...]

PR #142, opened 2020-11-13
Add markdown documentation for python-MAGNET interfacing

### LINKED ISSUE #123: Add markdown documentation of Python - MagNet critical information
PR #122 identifies two **Problems Encountered** and then describes solutions to these problems. Please write up these solutions in markdown documentation within the repository. I would like you to do this by creating a README.md file in `python/tools/magnet/`

Bharat-Ramadas:
# Description
This pull request captures the information provided in the `Problems Encountered` section of PR #122 description


python/tools/magnet/emach/README.md
```diff
@@ -0,0 +1,20 @@
+# Desciption
+
+This is a document aimed at helping users set up their systems in order to interface python with MAGNET
+
+- In order for Python to communicate with MAGNET, we need to import the pywin32 to call ActiveX compliant programs as COM automation servers. The pywin32 module comes with Anaconda3 installation and can be imported with the command `from win32com.client import DispatchEx`
```
ngadiyar93: @Bharat-Ramadas Probably changing sentences to third person makes it more consistent with rest of the documentation in eMach?
i.e, as an example, changing `In order for Python to communicate with MAGNET, we need to import the pywin32 to call ActiveX compliant programs as COM automation servers`  to `In order for Python to communicate with MAGNET, the pywin32 needs to be imported to call ActiveX compliant programs as COM automation servers` or equivalent

You could use this as an example: https://github.com/Severson-Group/eMach/blob/develop/README.md

ngadiyar93 (changes requested):
Nice work Bharat. Please see my comments.

python/tools/magnet/emach/README.md
Bharat-Ramadas: @ngadiyar93 Thanks for the feedback! I'll update the markdown to more closely match the example you've shared.

Bharat-Ramadas: @ngadiyar93 Please check the updated file and let me know your thoughts on the same.

python/tools/magnet/emach/README.md
```diff
@@ -0,0 +1,37 @@
+# Desciption
+
+This document is aimed at helping users write scripts for MAGNET in python.
+
+# Setup
+
+- In order for python to communicate with MAGNET, the pywin32 module has to be imported to call ActiveX compliant programs as Component Object Model (COM) automation servers. The pywin32 module comes with the Anaconda3 installation and can be imported with the command `from win32com.client import DispatchEx`
```
ngadiyar93: Where do I run this command- `from win32com.client import DispatchEx`?
Should the user run this in Anaconda Prompt? Or should this be added somewhere in the script? Or does it work either way?
May be a good idea to make it explicit?.

python/tools/magnet/emach/README.md
```diff
@@ -0,0 +1,37 @@
...
+- If the importing of the module raises an error flag, please make sure that the path of the Anaconda3 site-packages has been added to the system's `Environment variables` as `PYTHONPATH`.
```
ngadiyar93: May be better to say `If an error occurs and the module import is unsuccessful, please make sure.......` ?

python/tools/magnet/emach/README.md
```diff
@@ -0,0 +1,37 @@
...
+   MN.Visible = True #Makes MAGNET window visible
+   Doc = MN.newDocument()
+   View = Doc.getView()
+   Sol = Doc.getSolution()
+   MNConsts = MN.getConstants() 
+```
+# Function Returns
+
+MAGNET APIs can be divided into 2 categories based on how the function's return value can be accessed:
+- Functions who's return values can be directly accessed by assigining the function call to a variable fall under the first category. `makeSimpleCoil` is an example of such a function. Its return value can be obtained in python directly with the following syntax: `coil = Doc.makeSimpleCoil(ProblemID, ArrayOfValues)`
```
ngadiyar93: I guess you mean to say "whose" and not "who's" ?

python/tools/magnet/emach/README.md
```diff
@@ -0,0 +1,37 @@
...
+    MN.processCommand("REDIM strArray(0)")
+    MN.processCommand("DIM pType")
+    MN.processCommand("pType = getDocument.getParameter(\"{}\", \"{}\", strArray)".format(path,parameter))
+    MN.processCommand('Call setVariant(0, strArray,"PYTHON")')    
+    param = MN.getVariant(0,"PYTHON")
+    MN.processCommand('Call setVariant(0, pType,"PYTHON")')    
+    param_type = MN.getVariant(0,"PYTHON")
+    return param, param_type
+```
+The function definition of MAGNET APIs can be obtained from MAGNET's Help Topics.
```
ngadiyar93: @Bharat-Ramadas Do you think the MagNet knowledgebase article with the Python scripting example would be useful? If so, could you also add a link to it?

ngadiyar93 (changes requested):
Nice Work Bharat!. Please see my comments.

elsevers (changes requested):
- [x] Clean up grammar and some capitalization (`Python` and `MagNet`)
- [x] Provide more narrative framing to make this an easier read at the start of each section
- [x] Revise purpose statement to clarify that this doc is intended for developers and make the language more direct.

elsevers:
Why is this in an `emach` subfolder? We don't want to have a folder in the repo called `emach`

elsevers (approved):

ngadiyar93 (approved):
Looks good to me!

PR #144, opened 2020-11-20
Add Markdown Document addressing Issue #109

### LINKED ISSUE #109: Identify Python language features for implementing SOLIDD programming methodologies
Need to figure out how to do SOLIDD in Python!

What features of the Python language are best-used for implementing the SOLIDD programming methodologies? 

Add markdown documentation with references on how to do SOLIDD in Python to the repository. We have identified the following items as requiring markdown documents:
- [x] using the `abc` module to create abstract base classes
- [x] providing [type hints](https://www.python.org/dev/peps/pep-0484/) in Python
- [x] approach to checking type of function arguments and actions to take (throw exception?) if arguments are not of the expected type.
 - [ ] is it possible to enforce that classes which implement abstract base classes must provide functions with the same signature as is specified in the abstract base class?

aimthiazz:
Add Markdown Document addressing Issue #109


elsevers:
@mwjohnson56 will provide links to arguments against this approach.

Bharat-Ramadas (commented):
I think it would be better if this file were placed within the `python` folder of eMach. We can make a master README.md file that links to other documents on SOLIDD in Python

elsevers:
How are we doing on this PR?

mwjohnson56:
This should be moved to KB on python code.

elsevers:
Instead of merging this into eMach, let's put it into our knowledgebase section on Python and object oriented programming. I suggest that you split this out as separate files that you link to from the object oriented programming section.

PR #150, opened 2020-12-11
Add python support for Dimensions 

### LINKED ISSUE #139: Add dimensions to Python implementation

### LINKED ISSUE #160: Implement unary operations in DimAngular classes
This should be done in the same manner as DimLinear

aimthiazz:
Issue #139 
- [x] Protected to_dimes and from_dimes
- [x] Change name for to_dimes and from_dimes
- [x] Take advantage of polymorphism to have complete to_dimes and from_dimes functions defined in your abstract base class DimLinear


elsevers (changes requested):
Please change: 

- [x] DimLinear.py misspells `_to_dimesionless` and `_from_dimesionless`
- [x] DimLinear/DimLinear should return a scalar
- [x] Need to move `_to_dimensionless` and `_from_dimensionless` to the abstract base class. Recommended implementation: use an abstract conversion_factor property in these functions that the child classes must implement

python/model_obj/Dimensions/dim_linear.py
```diff
@@ -0,0 +1,85 @@
...
+    @abstractmethod
+    def to_inch(self):
+        pass
+
+    @property
+    @abstractmethod
+    def conversion_factor(self):
+        pass
+    
+    @abstractmethod
```
Bharat-Ramadas: Are `_to_dimensionless `and  `_from_dimensionless `supposed to be `abstractmethods`?

python/model_obj/Dimensions/main.py
```diff
@@ -0,0 +1,93 @@
+from dim_millimeter import DimMillimeter
```
Bharat-Ramadas: I think it would be better to have the `main.py` file in the `examples `folder

python/model_obj/Dimensions/dim_angular.py
```diff
@@ -0,0 +1,28 @@
+from dim_base import DimBase
```
Bharat-Ramadas: Please change `model_obj` location to be within the `emach` module

aimthiazz: Moved to `model_obj` location

python/model_obj/Dimensions/main.py
aimthiazz: Moved to `examples`  folder

python/model_obj/Dimensions/dim_linear.py
aimthiazz: Since ` to_millimeter()` and `to_inch()` are abstract method, I kept `_to_dimensionsless()` and `_from_dimensionless()` as abstractmethod.

Bharat-Ramadas (dismissed):
Looks good to me.

python/emach/model_obj/Dimensions/dim_angular.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class DimAngular(DimBase, ABC):
+    def __new__(cls, value):
+        return DimBase.__new__(cls, value)
+
+    def __init__(self, data):
+        DimBase.__init__(self, data)
+        
+    def __neg__(self):
```
aimthiazz, line 13: #160 Addressed

mwjohnson56 (dismissed):
Everything looks good here are ready to merge. I tested the code and it worked on my system. I have one small addition that I would like to be added as a new issue which is to utilize the `__repr__()` and `__str__()` functions to have the Dimensions objects output their units when displayed in the console or printed. See following code for the `DimInch` object.

` def __str__(self):
         return self.__repr__() `
`def __repr__(self):
        return float.__repr__(self) + ' [in]' `

python/emach/examples/example_dimension_unit_test.py
```diff
@@ -0,0 +1,93 @@
+# add the directory immediately above this file's directory to path for module import
```
elsevers, line 1: @aimthiazz I get a path error when trying to run this file.  Is this expected? How can I fix it?

image

python/emach/model_obj/Dimensions/dim_base.py
```diff
@@ -0,0 +1,16 @@
+
+class DimBase(float):
+    def __new__(cls, value):
+        return float.__new__(cls, value)
+    def __init__(self,data):
+        self.data = float(data)
+
+
```
elsevers: It seems to me that `DimBase` should have house these methods:

```python
    @property
    @abstractmethod
    def conversion_factor(self):
        pass
        
    @abstractmethod
    def _to_dimensionless(self):
        return float(self.data*self.conversion_factor)

    @abstractmethod    
    def _from_dimensionless(self):
        x = self.result/self.conversion_factor
        return self.__class__(x)
```

Do `_from_dimensionless` and `_to_dimensionless` need to be abstract or can they be concrete at the base class level?

python/emach/model_obj/Dimensions/dim_linear.py
```diff
@@ -0,0 +1,85 @@
...
+        pass
+    
+    @abstractmethod
+    def _to_dimensionless(self):
+        return float(self.data*self.conversion_factor)
+
+    @abstractmethod    
+    def _from_dimensionless(self):
+        x = self.result/self.conversion_factor
+        return self.__class__(x)
```
elsevers: I am confused about these functions being decorated with `@abstractmethod`. I would expect that this means that the classes inheriting `dim_linear` need to implement these methods. However, I can see that `dim_inch` and `dim_millimeter` do not implement these methods.  Rather, the methods appear to be fully defined here in `dim_linear`.

Can you explain?

python/emach/model_obj/Dimensions/dim_linear.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __pos__(self):
+        if self < 0:
+            return self * -1
+        else:
+            return self
+
+    @abstractmethod
+    def to_millimeter(self):
+        pass
```
elsevers: `dim_millimeter` and `dim_inch` appear to be implementing `to_mm` and not `to_millimeter`...

python/emach/model_obj/Dimensions/dim_linear.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @abstractmethod
+    def to_inch(self):
+        pass
```
elsevers: Do we have any need for these `to_inch` and `to_millimeter` functions any more?

elsevers (changes requested):

elsevers:
Let's remove the functions

`to_inch`
`to_millimeter` (and `to_mm`)
`to_degree`
`to_radian`

This was an old way of thinking and we are doing better now by using the to/from dimensionless approach

python/emach/model_obj/Dimensions/dim_linear.py
```diff
@@ -0,0 +1,63 @@
+from .dim_base import DimBase
+from abc import abstractmethod, ABC
+
+
+class DimLinear(DimBase, ABC):
```
Bharat-Ramadas: Does `DimLinear `need to inherit `ABC`?

python/emach/model_obj/Dimensions/dim_degree.py
```diff
@@ -0,0 +1,13 @@
...
+__all__ = ['DimDegree']
+
+class DimDegree(DimAngular):
+    def __new__(cls, value):
+        return DimAngular.__new__(cls, value)
+
+
+    @property
+    def conversion_factor(self):
+        return 3.14/180
```
Bharat-Ramadas: Its might be better to use `np.pi` here instead of 3.14

python/emach/examples/example_dimension_unit_test.py
Bharat-Ramadas: This error is happening because of a weird Git quirk. Git does not recognize letter case changes for files and folders. As a result, even though we changed the name of  `Dimensions`  to `dimensions` (first letter case change), the change has not been recognized by Git. However, Python is case sensitive and sees `Dimensions` and `dimensions` as different modules. This causes the error when someone tries to run the example file. 
You can fix this error by simply renaming the `Dimensions` folder within `model_obj` to `dimensions`.

python/emach/model_obj/Dimensions/dim_degree.py
aimthiazz: I wanted to avoid importing the NumPy library just for this file. I am circumventing the issue by increasing the significant digits of pi, i.e, Using `3.14159265359` instead of `3.14`

python/emach/examples/example_dimension_unit_test.py
aimthiazz: I resolved this issue by renaming the folder from `dimensions` to `dimension`, then back to `dimensions`.  Git has recognized this and changed the folder name to `dimensions` instead of `Dimensions`

Bharat-Ramadas (approved):
Looks good!

mwjohnson56 (approved):
Everything looks good, the required changes to the `dim_angular`, `dim_radian` and `dim_degree` have all been implemented in the same style as the `dim_linear` classes.

python/emach/model_obj/dimensions/dim_linear.py
```diff
@@ -0,0 +1,63 @@
...
+    def __sub__(self, other):
+        sub = self._to_dimensionless() - other._to_dimensionless()
+        self.result = sub
+        return type(self)._from_dimensionless(self)
+
+    def __mul__(self, other):
+        if (isinstance(self, DimLinear) and isinstance(other, DimLinear)):
+            raise Exception('Multiplication Not valid')
+
+        if isinstance(self, DimLinear):
```
elsevers, line 26: Is this check necessary? (Doesn't self have to be of type `DimLinear` since it is in the `DimLinear` class?

I also don't like it because this check implies that there is a path through this function that returns nothing (i.e., if somehow the object type is not an instance of `DimLinear` [which is impossible])

python/emach/model_obj/dimensions/dim_linear.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class DimLinear(DimBase):
+    def __new__(cls, value):
+        return DimBase.__new__(cls, value)
+
+    def __init__(self, data):
+        DimBase.__init__(self, data)
+
+    def __add__(self, other):
```
elsevers, line 12: I think we should actually just move all of these unary operations from `DimLinear` to `DimBase` so that they will work for angular as well as linear dimensions.

elsevers (changes requested):
Thanks @aimthiazz. Looks good!

See my detailed comments. My primary comment is that the unary operations of `DimLinear` could just as well apply to any dimension and so they should reside in the base class `DimBase`

elsevers (approved):
I am going to turn my revision requests into an issue and approve this code as-is, as it does implement the original request successfully. Nice job.

PR #159, opened 2020-12-21
Add hollow cylinder cross-section drawing capability 

### LINKED ISSUE #140: Add Location2D to Python implementation

### LINKED ISSUE #141: Add hollow cylinder cross-section to Python implementation
Add `crossects/hollowCylinder` capability to the Python support branch.  This will require you to have issues #139 and #140 implemented.

Please demonstrate that the follow script works (create an `eMach/python/examples` folder that holds a python version of this script):

```matlab

clc
clear

DRAW_MAGNET = 1;
DRAW_TIKZ   = 0;

%% Define cross sections

hollowCylinder1 = CrossSectHollowCylinder( ...
        'name', 'hollowCylinder1', ...
        'dim_d_a', DimMillimeter(4), ...
        'dim_r_o', DimMillimeter(80), ...
        'location', Location2D( ...
            'anchor_xy', DimMillimeter([0,0]), ...
            'theta', DimDegree(0).toRadians() ...
        ) ...
        );
    
%% Create MagNet tool
toolMn = MagNet();
toolMn.open(0,0,true);
toolMn.setDefaultLengthUnit('millimeters', false);


%% Draw the cylinder
hollowCylinder1.draw(toolMn)
toolMn.viewAll();
```

Bharat-Ramadas:
# Description

This PR adds all the functionality necessary for drawing cross-sections in python. This include `Location2D`, `CrossSectBase` and `CrossSectToken`, `HollowCylinder` class has been implemented and an example associated with the same as well.

# Related Issues
#140  #141


aimthiazz (dismissed):

python/emach/model_obj/cross_sects.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
```
ASHAD1: I think we need to use separate cross_section files for every component. So this should be renamed something like `cross_sect_hollow_cylinder.py`

python/emach/examples/example_hollow_cylinder_cross.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Fri Nov 13 01:02:08 2020
```
ngadiyar93: Minor comment: I am not a Python expert and unsure if we need all of these comments with date, author name etc? Maybe we want to keep it consistent with the Matlab style and not have any of that?

I know Spyder puts this in by default but it can be [changed](https://stackoverflow.com/questions/47619939/how-to-change-the-default-module-docstring-in-spyder).

python/emach/model_obj/__init__.py
```diff
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 """
-Created on Mon Feb  1 19:55:59 2021
+Created on Fri Nov 13 17:03:51 2020
```
ngadiyar93, line 1: Same comment about the date / author info.

python/emach/model_obj/cross_sects/cross_sect_base.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Thu Nov 12 23:22:49 2020
+
```
ngadiyar93, line 1: Same comment about the date / author info.

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
```diff
@@ -0,0 +1,94 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sat Mar  6 16:12:27 2021
+
```
ngadiyar93: Same comment about the date / author info.

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/README.md
```diff
@@ -0,0 +1,3 @@
+# Hollow Cylinder Cross Section
+
+<img src="./CrossSectHollowCylinder.svg" width="500" height="500" />
```
ngadiyar93: There seems to be some issue with the line color / fill color for the label `d_a`

aimthiazz (dismissed):
Works on my PC.  Approved.

ngadiyar93 (dismissed):

aimthiazz (dismissed):

python/emach/model_obj/cross_sects/cross_sect_base.py
```diff
@@ -0,0 +1,39 @@
...
+from abc import ABC, abstractmethod
+from typing import List
+
+class CrossSectBase(ABC):
```
elsevers, line 8: This class is missing the properties `name` and `location` as well as the method `clone()` that are provided in the MATLAB implementation. Is there a reason for not providing these in the abstract base class or where they just forgotten (in which case, please add)?

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+            origin
+
+        Returns
+        -------
+        None
+        '''
+        self.__name = name;
+        self.__dim_t = dim_t;
+        self.__dim_r_o = dim_r_o;
+        self.__location = location;
```
elsevers: In the MATLAB implementation `__name` and `__location` reside in the abstract base class. Seems like we should do the same here because that is part of the contract that a cross_sect must fulfill

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
```diff
@@ -0,0 +1,88 @@
+
+import numpy as np
+
+from ..cross_sect_base import CrossSectBase, CrossSectToken
+
+__all__ = ['CrossSectHollowCylinder']
+class CrossSectHollowCylinder(CrossSectBase):
+    
+    def __init__(self, name: str, dim_t: 'DimLinear', dim_r_o: 'DimLinear', \
```
elsevers: I see that your initializer has each dimension listed explicitly as an argument. The seems reasonable for simple geometry, like the hollow cylinder, but not great for complex geometry where there are a low of parameters (i.e. [inner rotor stator](https://github.com/Severson-Group/eMach/tree/develop/model_obj/crosssects/innerRotorStator)).  

In MATLAB, we used `varargin` to address this, and then accepted key value pairs, like a dictionary. 

What is the best practice for python? Is there style guidance on how to avoid absurdly length lists of arguments in class initializers and functions?

elsevers (changes requested):
Thanks @Bharat-Ramadas -- I think this is looking pretty good. I have a few requests entered in below. Also, I have this question / request that I have also put into slack:

> Tell me about the folder name `cross_sect_hollow_cylinder` -- it seems quite redundant to have `crosssects\cross_sect_xxxx`. Can you propose something to remove some of the redundancy?

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
Bharat-Ramadas: In Python, the recommended approach is to use keyword arguments or `**kwargs`. By doing this, you can pass in any number of named arguments and the arguments will be stored as a dictionary of key-value pairs within the function. This [link](https://stackabuse.com/variable-length-arguments-in-python-with-args-and-kwargs/) provides more information on keyword arguments in python

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
Bharat-Ramadas: In Python, we cannot define abstract properties, we can only define abstract methods. Specifying `name` and `location` as abstract methods might cause confusion which is why I decided not to place them within the abstract base class

python/emach/model_obj/cross_sects/cross_sect_base.py
Bharat-Ramadas: In Python, we cannot define abstract properties, we can only define abstract methods. Specifying name and location as abstract methods might cause confusion which is why I decided not to place them within the abstract base class. `clone( )` is something I forgot to implement, so I'll work on that.

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
elsevers: @Bharat-Ramadas in this case, these properties do not need to be abstract. They are repeated functionality which we want to guarantee that every class which inherits the base class has. Sure seems to me that they belong in the base class. Is it possible to make use of the base class initializer to ensure they are always set up properly?

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
elsevers: Great. Do it!

elsevers: Looks like if you are going to do this, you should also bring over the property validation architecture that the MATLAB cross-sects use. (`validateProps` and `createProps`)

Bharat-Ramadas: I've implemented both `validateProps `and `createProps `(`_create_attr ` and `_validate_attr `in Python)

python/emach/model_obj/cross_sects/cross_sect_hollow_cylinder/__init__.py
Bharat-Ramadas: I've implemented the validateProps Matlab equivalent of `name` and `location` in `cross_sect_base`. `createProps` has now been moved to reside in the base class as it seems that the same function is repeated in each `cross_sect`

aimthiazz (dismissed):
Works on my PC. Looks good.

python/emach/examples/example_hollow_cylinder_cross.py
```diff
@@ -0,0 +1,36 @@
...
+import numpy as np
+
+x = mo.DimMillimeter(4)
+y = mo.DimMillimeter(80)
+z = mo.DimMillimeter(40)
+
+# create hollowcylinder crossection object
+hollowCylinder1 = mo.cross_sects.CrossSectHollowCylinder(name = 'hollowCylinder1',
+                                                location = mo.Location2D(),
+                                                dim_t = x, dim_r_o = y)
```
ASHAD1, line 16: Looks like in this example you are not extruding and assigning materials, do we need this in this example?

Bharat-Ramadas: @ASHAD1 Yeah, that will come after we implement `eMach` `components`. Right now, this is just to test the cross-sect functionality. We haven't gotten to that point yet in the python implementation.

ASHAD1: Okay, I see now. It's drawing the cross-section, not components. I will go through this PR in detail again. Thanks.

aimthiazz (dismissed):

python/emach/model_obj/cross_sects/hollow_cylinder/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+    def dim_r_o(self):
+        return self._dim_r_o
+    
+    @property
+    def name(self):
+        return self._name
+    
+    @property
+    def location(self):
+        return self._location
```
elsevers: It seems to me that `name` and `location` belong in the base class to make your code DRY because all cross sects have these properties.

python/emach/model_obj/location_2d.py
```diff
@@ -0,0 +1,84 @@
...
+    
+    @property
+    def theta(self):
+        return self.__theta
+    
+    @property
+    def rot(self):
+        return self.__rot
+    
+    def trans_coord(self, coords, add_theta = None):
```
elsevers: Please name this function `transform_coords`

elsevers (changes requested):
Looks great! I have two small change requests (see my comments on your code). Address these, and I'll approve this.

python/emach/model_obj/cross_sects/hollow_cylinder/__init__.py
Bharat-Ramadas: Done!

python/emach/model_obj/location_2d.py
Bharat-Ramadas: Sure thing

aimthiazz (approved):

elsevers (approved):
Looks good. Thanks @Bharat-Ramadas

PR #166, opened 2021-02-21
Add TokenDraw Class 

### LINKED ISSUE #165: Create TokenDraw Class to hold eMach segments data

Bharat-Ramadas:
# Description
This PR adds the `TokenDraw ` class and supporting changes in the magnet tool module.

# Related Issues
#165

### DIFF
--- a/python/emach/tools/magnet/magnet.py
+++ b/python/emach/tools/magnet/magnet.py
@@ -1,9 +1,3 @@
-# -*- coding: utf-8 -*-
-"""
-Created on Fri Oct 16 16:51:18 2020
-
-@author: Bharat
-"""
 from win32com.client import DispatchEx
 
 from ..tool_abc import toolabc as abc
--- a/python/emach/tools/magnet/magnet.py
+++ b/python/emach/tools/magnet/magnet.py
@@ -5,8 +5,9 @@ Created on Fri Oct 16 16:51:18 2020
 @author: Bharat
 """
 from win32com.client import DispatchEx
-from ..tool_abc import toolabc as abc
 
+from ..tool_abc import toolabc as abc
+from ..token_draw import TokenDraw
 from . import document
 from .document import *
 
@@ -22,7 +23,11 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
     """
 
     def __init__(self):
-        pass
+        self.disp_ex = None
+        self.doc = None
+        self.view = None
+        self.sol = None
+        self.consts = None
 
     def open(self, filename=None, visible=False):
         """ opens a new MAGNET session and assigns variables neccessary for further
@@ -64,11 +69,13 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
             0 if successful 1 if failed.
 
         """
-        try:
-            ret = self.view.newLine(startxy[0], startxy[1], endxy[0], endxy[1])
-            return ret
-        except:
-            raise TypeError
+        ret = self.view.newLine(startxy[0], startxy[1], endxy[0], endxy[1])
+        
+        self.disp_ex.setVariant(0, 'line', 'python')
+        line = self.disp_ex.getVariant(0, 'python');
+        
+        return TokenDraw(line,1)
+
 
     def draw_arc(self, centrexy, startxy, endxy):
         """
@@ -85,14 +92,17 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
 
         Returns
         -------
-        ret : int
-            0 if successful 1 if failed.
+        ret : TokenDraw object
 
         """
         ret = self.view.newArc(
             centrexy[0], centrexy[1], startxy[0], startxy[1], endxy[0], endxy[1]
         )
-        return ret
+        
+        self.disp_ex.setVariant(0, 'arc', 'python')
+        arc = self.disp_ex.getVariant(0, 'python');
+        
+        return TokenDraw(arc,1)
 
     def select(self):
         pass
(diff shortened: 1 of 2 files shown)


python/emach/tools/token_draw.py
```diff
@@ -0,0 +1,34 @@
...
+@author: Bharat
+"""
+
+class TokenDraw():
+    '''
+    This class acts as a wrapper of information generated when drawing a line 
+    or an arc using emach
+    '''
+    
+    def __init__(self, segment_indices: any, segment_type: int) -> None :
```
aimthiazz, line 14: typehint `->any`

aimthiazz (dismissed):
Looks good.

ngadiyar93 (dismissed):
Looks good to me.

python/emach/tools/token_draw.py
```diff
@@ -0,0 +1,34 @@
...
+            This specifies whether the segment drawn is a line or an arc. 1 
+            represents arcs and 0 represents lines
+
+        Returns
+        -------
+        None.
+
+        '''
+        self.segment_indices = segment_indices
+        self.segment_type = segment_type
```
elsevers: Please match variable names to the MATLAB code.

python/emach/tools/token_draw.py
```diff
@@ -0,0 +1,34 @@
...
+
+@author: Bharat
+"""
+
+class TokenDraw():
+    '''
+    This class acts as a wrapper of information generated when drawing a line 
+    or an arc using emach
+    '''
+
```
elsevers, line 7: Make properties read only

elsevers (changes requested):

aimthiazz (dismissed):

python/emach/tools/token_draw.py
```diff
@@ -0,0 +1,43 @@
...
+            
+        segment_type : int
+            This specifies whether the segment drawn is a line or an arc. 1 
+            represents arcs and 0 represents lines
+
+        Returns
+        -------
+        None.
+
+        '''
```
elsevers: There parameter names in this comment block do not match the arguments in your initializer. 

Also, is it true that an initializer returns nothing? It seems to me that it returns a `TokenDraw` object.

Finally, why is there so much vertical white space in this comment block? Are you following a certain Python style, or is this a mistake?

elsevers (changes requested):

python/emach/tools/token_draw.py
Bharat-Ramadas: The docustring has been revised to reflect the function arguments

The initializer does not return anything in Python. It is the constructor that returns the instance of the class ( `__new__` ). Most of the time the constructor does not have to be explicitly declared in Python (however, we have made use of `__new__` in the implementation of `dimensions`).

The functions docstrings outline gets generated automatically when using Spyder. I used this as the default documentation and made changes on top of it where required. This resulted in the large vertical while space in the comment. Some of the vertical space was provided for descriptions and other are to segregate between different sections of the docustring. I looked up some examples and [PEP257](https://www.python.org/dev/peps/pep-0257/) found that there is no style specification on vertical white space.

python/emach/tools/magnet/magnet.py
```diff
@@ -118,49 +128,36 @@ def prepare_section(self, inner_coord):
             [self.consts.infoSliceSurface],
         )
 
-    def extrude(self, name, material, depth, token=None):
+    def extrude(self, name, material, depth):
```
elsevers: Why are you removing the token from this function signature? 

All classes implementing `MakerExtrudeBase` need to be able to receive a token here

python/emach/tools/magnet/magnet.py
```diff
@@ -118,49 +128,36 @@ def prepare_section(self, inner_coord):
...
-                or self.consts.infoMakeComponentRemoveVertices,
+                self.consts.infoMakeComponentRemoveVertices
             )
-        else:
-            ret = self.View.makeComponentInALine(depth, name, "Name=" + material, token)
+
         return ret
 
-    def revolve(self, name, material, center, axis, angle, token=None):
+    def revolve(self, name, material, center, axis, angle):
```
elsevers: Why is `token` being removed here? This does not match the MATLAB implementation

elsevers (changes requested):
@Bharat-Ramadas your implementation of `TokenDraw` looks good to me, but I am confused about your changes to `magnet.py`. Why are you removing the tokens on extrude and revolve? This doesn't match the MATLAB eMach implementation.

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: In our last meeting, the point was raised that `token` should be removed from both the Matlab and Python implementation. Even though the `MagNet extrude` function has an argument `token`, it is not used anywhere within the function. Also, the function signature provided in the base class `MakerExtrudeBase` (Matlab) does not include `token`
https://github.com/Severson-Group/eMach/blob/f8bd845a7275ece4f4618324923e5ff66876eb77/tools/makerBase/MakerExtrudeBase.m#L12

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: In our last meeting, the point was raised that `token` should be removed from both the Matlab and Python implementation. Even though the `MagNet revolve` function has an argument `token`, it is not used anywhere within the function. Also, the function signature provided in the base class `MakerRevolveBase` (Matlab) does not include `token`
https://github.com/Severson-Group/eMach/blob/f8bd845a7275ece4f4618324923e5ff66876eb77/tools/makerBase/MakerRevolveBase.m#L13

python/emach/tools/magnet/magnet.py
elsevers: @Bharat-Ramadas, I don't remember this being the outcome of that discussion. Is there a PR or issue you can point me to that will refresh my memory?

These tokens are not actually used by MagNet / JMAG right now, but it is highly likely a future tool may require this information. To see why this is true, follow the flow of information in the `run()` function of `MakeExtrude.m` and `MakeRevolve.m`:

https://github.com/Severson-Group/eMach/blob/f8bd845a7275ece4f4618324923e5ff66876eb77/model_obj/makeSolid/MakeExtrude.m#L17-L34

Written in psuedocode, this looks like this:

```
1) csToken = drawer.drawTheCrossSection()
2) prepareToken = maker.prepareSection(csToken)
3) extrudeToken = maker.extrudeSection(prepareToken)

tokenMake = TokenMake(csToken, prepareToken, extrudeToken)
```
Do you see how `tokenMake` now contains all custom information generated my the tool (or maker)? This is returned to the eMach user when they call `tokenMake = component.make(drawer, maker)`. The user can then use this information for more advanced/custom interfacing to the tool. I.e., this is likely to be important to code that is actually solving for fields in the machine.

Make sense?

elsevers (changes requested):

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: Yup, you've convinced me. I've reverted the functions back to their previous argument signature.

aimthiazz (dismissed):
Looks good

python/emach/tools/magnet/magnet.py
```diff
@@ -5,8 +5,9 @@
 @author: Bharat
 """
```
ngadiyar93: All the python files seem to have this header with author name, date created etc. Do we want to keep that? 
It seems to be an artifact from Spyder.....

ngadiyar93 (commented):
I have a comment that I have indicated. Else, looks good to me.

elsevers (dismissed):
I am approving this based on `token_draw`. 

The MagNet tool class needs further work to be compliant with the latest spec of the MATLAB version: `prepare_section` should return a token, units of dimensions are not handled, and the type hints are needed.

elsevers:
I created issue #183 to track the improvements needed to the MagNet class

elsevers (approved):

ngadiyar93 (approved):

PR #176, opened 2021-03-06
Fix Issue #173 , #174  and #175

### LINKED ISSUE #173: Move unary operations to DimBase so that they also apply to DimAngular
PR #150 added dimensions and created unary operations for linear dimensions. These operations also should apply to angular dimensions. 

Please move the unary operations to the `DimBase` class so that they work for both linear and radial dimensions.

https://github.com/Severson-Group/eMach/blob/beb4fc720a5c1c3abd741eadc791706c0b621e63/python/emach/model_obj/dimensions/dim_linear.py#L12-L61

### LINKED ISSUE #174: Clean up DimLinear code to remove ambiguous return paths
Several functions in the `DimLinear` class have conditional paths that are not complete. For example, see the code below:

https://github.com/Severson-Group/eMach/blob/beb4fc720a5c1c3abd741eadc791706c0b621e63/python/emach/model_obj/dimensions/dim_linear.py#L22-L31

There is no `else` for the last `if`, and so it appears as though it is possible for the program to exit this function without returning a value. This add potential for errors in the code. Please clean up.

Further, I question whether it is necessary to check `isinstance(self, DimLinear)` -- since this code is appearing in the `DimLinear` class, wouldn't `self` always be of type `DimLinear`?

Please clean up these functions:
- [x] `__mul__`
- [x] `__rmul__`
- [x] `__truediv__`
- [x] `__rtruediv__`

### LINKED ISSUE #175: Python dimensions should raise exception for squared operation
The following code should throw an exception, but it does not:

```python
inchObject = DimInch(1)
inchObject**2
```

Please fix.

aimthiazz:
This PR will address issues #173, #174, and #175.

### DIFF
--- a/python/emach/examples/example_dimension_unit_test.py
+++ b/python/emach/examples/example_dimension_unit_test.py
@@ -85,6 +85,8 @@ print(DimDegree(radObject))
 
 print('Angle Unary operation')
 print(-degObject)
+negDegObject = DimDegree(-180)
+print(-negDegObject)
 
 
 
--- a/python/emach/examples/example_dimension_unit_test.py
+++ b/python/emach/examples/example_dimension_unit_test.py
@@ -71,7 +71,8 @@ print(type(inchObject/5))
 
 ###### Uncomment this to check exception #######
 # print('********Power Demo*****')
-# print(inchObject**2)
+#print(inchObject**2)
+#print(2**inchObject)
 # print(inchObject**milObject)
 
 print('*******Angle Demo******')
--- a/python/emach/examples/example_dimension_unit_test.py
+++ b/python/emach/examples/example_dimension_unit_test.py
@@ -69,6 +69,11 @@ print(type(inchObject/5))
 # print(5/inchObject)
 # print(type(inchObject/5))
 
+###### Uncomment this to check exception #######
+# print('********Power Demo*****')
+# print(inchObject**2)
+# print(inchObject**milObject)
+
 print('*******Angle Demo******')
 degObject = DimDegree(180)
 print(type(degObject))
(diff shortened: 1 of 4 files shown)


elsevers (changes requested):
@aimthiazz this seems to implement issue #174 and #175, but it does not fully implement #173.

`DimLinear` still contains functions that need to be moved to `DimBase` in order to implement #173:

```python
def __add__(self, other):
        add = self._to_dimensionless() + other._to_dimensionless()
        self.result = add
        return type(self)._from_dimensionless(self)

    def __sub__(self, other):
        sub = self._to_dimensionless() - other._to_dimensionless()
        self.result = sub
        return type(self)._from_dimensionless(self)

    def __mul__(self, other):
        if (isinstance(self, DimLinear) and isinstance(other, DimLinear)):
            raise Exception('Multiplication Not valid')

        else:
            mul = (other) * (self._to_dimensionless())
            self.result = mul 
            return type(self)._from_dimensionless(self)

    def __rmul__(self, other):
        if (isinstance(self, DimLinear) and isinstance(other, DimLinear)):
            raise Exception('Multiplication Not valid')

        else:
            mul = (other) * (self._to_dimensionless())
            self.result = mul
            return type(self)._from_dimensionless(self)


    def __truediv__(self, other):
        if (isinstance(other,DimLinear)):
            div = self._to_dimensionless() / other._to_dimensionless()
            return div
        else:
            div = (self._to_dimensionless()) / (other)
            self.result = div
            return type(self)._from_dimensionless(self)

    def __rtruediv__(self, other):
        raise Exception('Division not valid')

```

Bharat-Ramadas (dismissed):
The example runs as expected

elsevers (approved):
Looks good. Thanks @aimthiazz

Bharat-Ramadas (approved):

PR #181, opened 2021-03-06
Update dimlinear child classes to address nested dimension issue

### LINKED ISSUE #180: Incorrect implementation of nested Dimensions
# Description
Nested implementation of Dimensions, both DimLinear and DimAngular does not work as intended. Considered the following example:
```python
inchObject = DimInch(1)
mmObject = DimMillimeter(inchObject )
print(mmObject)
```
In the above scenario, the expected output is that `mmObject` will equal 25.4 as the `conversion factor` of `DimInch` is 25.4 while that of `DimMillimeter` is 1. However, in the current Python implementation, this results in 1 or the same value as what the `DimInch` object was declared with.

# Recommended Solution
In the constructor of each `DimLinear` and `DimAngular` child class, an addition check should be in place to check if the value being passed is a `DimLinear` or `DimAngluar` object. If so, the value passed should first be scaled with the corresponding class's conversion factor before return. This is similar to Matlab implementation.
https://github.com/Severson-Group/eMach/blob/435ab7c73fcdc3c981f93c6779bdecc489bfd269/model_obj/dimensions/DimMillimeter.m#L10

aimthiazz:
Addressing Issue #180

### DIFF
--- a/python/emach/model_obj/dimensions/dim_linear.py
+++ b/python/emach/model_obj/dimensions/dim_linear.py
@@ -5,42 +5,4 @@ class DimLinear(DimBase):
     def __new__(cls, value):
         return DimBase.__new__(cls, value)
 
-    def __add__(self, other):
-        add = self._to_dimensionless() + other._to_dimensionless()
-        result = type(self)(add)
-        return type(self)._from_dimensionless(result)
 
-    def __sub__(self, other):
-        sub = self._to_dimensionless() - other._to_dimensionless()
-        result = type(self)(sub)
-        return type(self)._from_dimensionless(result)
-
-    def __mul__(self, other):
-        if isinstance(self, DimLinear) and isinstance(other, DimLinear):
-            raise Exception('Multiplication Not valid')
-
-        if isinstance(self, DimLinear):
-            mul = other * (self._to_dimensionless())
-            result = type(self)(mul)
-            return type(self)._from_dimensionless(result)
-
-    def __rmul__(self, other):
-        if isinstance(self, DimLinear) and isinstance(other, DimLinear):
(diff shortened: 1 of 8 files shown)


Bharat-Ramadas (dismissed):
Everything is working as expected on my PC. Good work @aimthiazz!

python/emach/model_obj/dimensions/dim_linear.py
```diff
@@ -1,63 +1,46 @@
...
 
-    def __init__(self, data):
-        DimBase.__init__(self, data)
-
     def __add__(self, other):
         add = self._to_dimensionless() + other._to_dimensionless()
-        self.result = add
-        return type(self)._from_dimensionless(self)
+        result = type(self)(add)
+        return type(self)._from_dimensionless(result)
```
elsevers: This needs to be cleaned up

Implement `_from_dimensionless` as a class method in `DimBase` that takes in a type of object to create and then a dimensionless float

elsevers (changes requested):

Bharat-Ramadas (commented):
Looks great @aimthiazz! Just one minor comment that I have is to rename `conversion_factor` to `_conversion_factor` so that users know they are not supposed to access it.

Bharat-Ramadas (dismissed):
Looks good

ngadiyar93:
@aimthiazz I haven't taken a close look at the changes, but it looks like you have few merge conflicts that need to be resolved.

aimthiazz:
Thanks, @ngadiyar93. I have resolved the conflict.

Bharat-Ramadas (approved):

elsevers (approved):
Looks good!

PR #182, opened 2021-03-07
Update MagNet MATLAB tool to properly handle the units of received dimensions

### LINKED ISSUE #153: MagNet tool needs to properly handle the units of received dimensions
To finish for:
- [x] MATLAB (@ngadiyar93 )
- [x] Python (@Bharat-Ramadas )

The `drawLine` and `drawArc` functions in the MagNet tool are directly taking received linear and angular dimensions and passing their value to the MagNet API without converting to appropriate units. This can cause problems if a caller of these functions uses an unexpected unit. It should not be the responsibility of the caller to guarantee that the unit matches MagNet's expectations. Rather, the MagNet tool class should convert the received values into the appropriate units before passing the value to the FEA software.

**Current code:**

https://github.com/Severson-Group/eMach/blob/a1f209d25dee4c8db9b9dae0ab14e819871a7c3e/tools/magnet/MagNet.m#L75-L85

https://github.com/Severson-Group/eMach/blob/a1f209d25dee4c8db9b9dae0ab14e819871a7c3e/tools/magnet/MagNet.m#L95-L105

**Suggested implementation:**
- create properties in the tool class that indicate which dimension class should be used for angular and linear dimensions. 
- in `drawLine` and `drawArc`, use `feval()` with these properties to fabricate appropriate dimension objects that have the expected units
- in `drawLine` and `drawArc`, pass the value of the fabricated dimension object to the FEA software

This has already partially been implemented via the `defaultLength` property. However, this implementation is not finished. Need to finish it.

https://github.com/Severson-Group/eMach/blob/a1f209d25dee4c8db9b9dae0ab14e819871a7c3e/tools/magnet/MagNet.m#L8-L14

ngadiyar93:
This PR updates the MagNet MATLAB tool to properly handle units of received dimensions.

1. `defaultLength` and `defaultAngle` properties have been created to set the drawing units in the tool. 
2. `drawLine`, `drawArc`, `extrude`, and `revolve` have been updated to use `feval` to fabricate appropriate dimension objects that have the expected units.
3. The value of the fabricated dimension object using `feval` is passed to the FEA software.

### DIFF
--- a/tools/magnet/MagNet.m
+++ b/tools/magnet/MagNet.m
@@ -169,7 +169,7 @@ classdef MagNet < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
             seltype = get(obj.consts,'InfoSetSelection');
             objcode = get(obj.consts,'infoSliceSurface');
             
-            %TO DO: how to deal with the units of the coordinate?? 
+            csToken.innerCoord = feval(obj.defaultLength, csToken.innerCoord); 
             mn_dv_selectat(obj.mn, csToken.innerCoord, seltype, objcode);
             new = 1;
         end
--- a/tools/magnet/MagNet.m
+++ b/tools/magnet/MagNet.m
@@ -10,7 +10,8 @@ classdef MagNet < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
         doc; % Document object
         view; %View object
         consts; %Program constants
-        defaultLength = 'dimMillimeter';
+        defaultLength = 'DimMillimeter';
+        defaultAngle  = 'DimDegree'
     end
     
     methods
@@ -77,7 +78,8 @@ classdef MagNet < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
             %   drawLine([start_x, _y], [end_x, _y]) draws a line
             %
             %   This is a wrapper for the Document::View::newLine function.
-  
+            startxy = feval(obj.defaultLength, startxy);
+            endxy = feval(obj.defaultLength, endxy);     
             invoke (obj.mn, 'processcommand', 'redim line(0)');
             invoke (obj.mn, 'processcommand', sprintf(...
                 'call getDocument.getView.newLine( %f, %f, %f, %f, line)', ...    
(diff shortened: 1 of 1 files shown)


Bharat-Ramadas (changes requested):
Good job @ngadiyar93! However, I think a couple more changes are required in order to make this PR complete. 

1. Inner coordinate being passed to prepareSection has to be converted to default units https://github.com/Severson-Group/eMach/blob/807817a9c554807befa319672fe1f05b114bb8cc/tools/magnet/MagNet.m#L173
2. setDefaultLengthUnit function should not have if-else statements within it. Instead the function should directly parse user input. https://github.com/Severson-Group/eMach/blob/807817a9c554807befa319672fe1f05b114bb8cc/tools/magnet/MagNet.m#L177-L206

ngadiyar93:
@Bharat-Ramadas Thanks, I have updated to address 1.

I am not very clear what you mean by `The function should directly parse user input`. Could you clarify?
The way it is setup right now, a user can call for example  `setDefaultLengthUnit(tool.mn, 'inches', 1)` How do you suggest I handle it without using the `if - else`?

Bharat-Ramadas:
@ngadiyar93 what I mean by that is instead of the function call being `setDefaultLengthUnit(tool.mn, 'inches', 1)`, it should be `setDefaultLengthUnit(tool.mn, 'DimInches', 1)` so that the function does not need to conduct any checks and modify user input.

ngadiyar93:
@Bharat-Ramadas The function will have to check / take action on user input even if I modify as suggested beacause:
The units are set in the FEA tool by `invoke(obj.doc, 'setDefaultLengthUnit', userUnit, makeAppDefault);` 
This expects `userUnit` to be a string that MagNet expects (`millimeters`/ `inches`) and not `DimMillimeter` / `DimInch` 

So, if I go with the route you suggested, I still need to have an if-else to get the appropriate string to pass on-to the FEA tool. So I don't see a clear advantage of changing it from the way it is. Or am I missing something here?

ngadiyar93:
@Bharat-Ramadas I have also reverted the change to the `Inner coordinate` being passed to `prepareSection`.
If I use `feval` MATLAB errors out saying `innerCoord` is a read-only property.

image

Bharat-Ramadas:
@ngadiyar93 I agree with your comments. The only advantage I see in passing `DimInch` instead of `inches` is that `DimInch` is a fixed object as far as eMach is concerned whereas `inches`, which is the argument passed in directly to the FEA tool is something that will vary with each tool (for eg: for JMAG we pass `SI Units`).  We can have a discussion on this during the group meeting. Everything else looks good!

Bharat-Ramadas (approved):
Looking good!

elsevers (approved):
I tested this for a few cases and it works great. Nice job @ngadiyar93

PR #184, opened 2021-03-08
Update MagNet tool to handle dimension and to revise application/file handling

### LINKED ISSUE #153: MagNet tool needs to properly handle the units of received dimensions

### LINKED ISSUE #172: Tools should use open and close for files, not the application instance
Revise the tools (MagNet and JMAG) so that:

1. `open` and `close` are used to open and close a project/file
2. The application itself is opened when the tool object instance is created (in the constructor)
3. Create a destructor method that is called when the tool object instance is destroyed. This destructor should exit the tool application. See [here](https://www.mathworks.com/help/matlab/matlab_oop/handle-class-destructors.html) for instructions on how to do this in MATLAB.
4. Create an example demonstrating the complete lifecycle of a tool. That is, create a tool object instance (see that the tool application opens), open a file (or create a new file), close the file, and then destroy the object instance and watch the tool application close)

Code to change:
1. The tool class (i.e. `MagNet.m`)
    1. `visible` moves from the `open()` function to the constructor
    2. `mn` moves from the `open()` function to the constructor
2. The `tool` abstract base class (update comments and function signature)

https://github.com/Severson-Group/eMach/blob/f8bd845a7275ece4f4618324923e5ff66876eb77/tools/magnet/MagNet.m#L17-L65

Bharat-Ramadas:
# Description
This PR implements the issue raised in #153 and #172 . The `MagNet` tool now converts all `dimensions` to the default setting before passing the values on to the FEA software. The tool also opens the application upon object `construction` and closes application upon calling the `destructor`. 

# Problems Encountered
In order for the `MagNet` tool to access the dimensions classes such as `DimMillimeter` and `DimInch`, importing those classes from the corresponding module is necessary. `Python`, however does not allow a user to import modules that are at or above the directory where the execution file resides. Therefore, it became necessary to move the example folder to one level higher in order to get the code working.

### DIFF
--- a/python/emach/tools/magnet/magnet.py
+++ b/python/emach/tools/magnet/magnet.py
@@ -22,6 +22,9 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
         self.view = None
         self.sol = None
         self.consts = None
+        self.default_length = None
+        self.default_angle = None
+        self.filename = None
         
     def __del__(self):
         self.mn.close(False)
@@ -35,6 +38,7 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
         
         if filename is str:
             self.doc = self.mn.openDocument(filename)
+            self.filename = filename
         else:
             self.doc = self.mn.newDocument()
         self.view = self.doc.getView()
@@ -46,9 +50,15 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
     def close(self):
         self.doc.close(False)
         
-    def save(self, path, filename):
-        filepath = path+filename
-        self.doc.save(filepath)
+    def save_as(self, filename):
+        self.filename = filename
+        self.save()
+        
+    def save(self):
+        if self.filename is str:
+            self.doc.save(self.filename)
+        else:
+            raise AttributeError('Unable to save file. Use the save_as() function')
         
     def draw_line(self, startxy, endxy):
         """
--- a/python/emach/tools/magnet/magnet.py
+++ b/python/emach/tools/magnet/magnet.py
@@ -15,22 +15,25 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
     """ A class to handle MagNet applications
     """
 
-    def __init__(self):
-        self.disp_ex = None
-        self.doc = None
+    def __init__(self, visible = False):
+        self.disp_ex = DispatchEx("MagNet.Application")
+        self.disp_ex.visible = visible 
         self.view = None
         self.sol = None
         self.consts = None
         self.default_length = 'DimMillimeter'
         self.default_angle = 'DimDegree'
+    
+    def __del__(self, filepath = None):
+        if filepath is str:
+            self.doc.save(filepath)
+        self.disp_ex.close(False)
 
-    def open(self, filename=None, visible=False):
+    def open(self, filename=None):
         """ opens a new MAGNET session and assigns variables neccessary for further
         operations
         """
 
-        self.disp_ex = DispatchEx("MagNet.Application")  # Opens a new MAGNET session
-        self.disp_ex.visible = visible  # Makes MAGNET window visible
         if filename is str:
             self.doc = self.disp_ex.openDocument(filename)
         else:
@@ -39,8 +42,10 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
         self.sol = self.doc.getSolution()
         self.consts = self.disp_ex.getConstants()  # Get MAGNET Constants
 
-    def close(self):
-        pass
+    def close(self, filepath = None):
+        if filepath is str:
+            self.doc.save(filepath)
+        self.view.close(False)
 
     def draw_line(self, startxy, endxy):
         """
--- a/python/emach/tools/magnet/magnet.py
+++ b/python/emach/tools/magnet/magnet.py
@@ -16,37 +16,40 @@ class MagNet(abc.ToolBase, abc.DrawerBase, abc.MakerExtrudeBase, abc.MakerRevolv
     """
 
     def __init__(self, visible = False):
-        self.disp_ex = DispatchEx("MagNet.Application")
-        self.disp_ex.visible = visible 
+        self.mn = DispatchEx("MagNet.Application")
+        self.mn.visible = visible 
+        self.doc = None
         self.view = None
         self.sol = None
         self.consts = None
-        self.default_length = 'DimMillimeter'
-        self.default_angle = 'DimDegree'
-    
-    def __del__(self, filepath = None):
-        if filepath is str:
-            self.doc.save(filepath)
-        self.disp_ex.close(False)
+        
+    def __del__(self):
+        self.mn.close(False)
 
-    def open(self, filename=None):
+    def open(self, filename=None, length_unit = 'DimMillimeter', angle_unit = 'DimDegree'):
         """ opens a new MAGNET session and assigns variables neccessary for further
         operations
         """
-
+        self.default_length = length_unit
+        self.default_angle = angle_unit
+        
         if filename is str:
-            self.doc = self.disp_ex.openDocument(filename)
+            self.doc = self.mn.openDocument(filename)
         else:
-            self.doc = self.disp_ex.newDocument()
+            self.doc = self.mn.newDocument()
         self.view = self.doc.getView()
         self.sol = self.doc.getSolution()
-        self.consts = self.disp_ex.getConstants()  # Get MAGNET Constants
-
-    def close(self, filepath = None):
-        if filepath is str:
-            self.doc.save(filepath)
-        self.view.close(False)
-
+        self.consts = self.mn.getConstants()  # Get MAGNET Constants
+        
+        self.set_default_length(length_unit, False)
+        
+    def close(self):
+        self.doc.close(False)
+        
+    def save(self, path, filename):
+        filepath = path+filename
+        self.doc.save(filepath)
+        
     def draw_line(self, startxy, endxy):
         """
         Draws a line in MAGNET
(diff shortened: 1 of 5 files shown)


aimthiazz (dismissed):
Ran Example, works in my PC

python/emach/tools/magnet/magnet.py
```diff
@@ -4,32 +4,36 @@
...
-    """ A class to represent a MAGNET file
+    """ A class to handle MagNet applications
     """
 
-    def __init__(self):
-        self.disp_ex = None
-        self.doc = None
+    def __init__(self, visible = False):
+        self.disp_ex = DispatchEx("MagNet.Application")
+        self.disp_ex.visible = visible
```
elsevers: Why are you calling the magnet instance `disp_ex` and not `mn`. In MATLAB we use `mn` -- it seems like we should be using that same variable name here unless there is a reason not to.

python/emach/tools/magnet/magnet.py
```diff
@@ -4,32 +4,36 @@
...
-        self.disp_ex = None
-        self.doc = None
+    def __init__(self, visible = False):
+        self.disp_ex = DispatchEx("MagNet.Application")
+        self.disp_ex.visible = visible 
         self.view = None
         self.sol = None
         self.consts = None
+        self.default_length = 'DimMillimeter'
+        self.default_angle = 'DimDegree'
```
elsevers: Your initializer isn't allowing you to specify `default_length` and `default_angle` (the MATLAB version does allow this to be specified). I suggest you add them as function arguments here with default values.

python/emach/tools/magnet/magnet.py
```diff
@@ -4,32 +4,36 @@
...
+    
+    def __del__(self, filepath = None):
+        if filepath is str:
+            self.doc.save(filepath)
```
elsevers: The MATLAB implementation does not have an optional argument for saving upon closing. At our last eMach meeting we agreed that was the behavior we wanted. Please remove that functionality from here (remove lines 28 and 29).

python/emach/tools/magnet/magnet.py
```diff
@@ -38,8 +42,10 @@ def open(self, filename=None, visible=False):
         self.sol = self.doc.getSolution()
```
elsevers, line 45: Interesting:
```python
self.sol = self.doc.getSolution()
```
I am unable to test this at the moment, but I am wondering -- does this cause an error if document does not have a solution? (I.e., this is a model that is not solved):
- *If it does throw an error,* we need to remove this
- *If it does not throw an error,* please add a review request on PR #186 for @ngadiyar93 to also get the solution.

python/emach/tools/magnet/magnet.py
```diff
@@ -38,8 +42,10 @@ def open(self, filename=None, visible=False):
...
         self.consts = self.disp_ex.getConstants()  # Get MAGNET Constants
```
elsevers: You must also set MagNet's default unit length based on the value of `default_length`

python/emach/tools/magnet/magnet.py
```diff
@@ -38,8 +42,10 @@ def open(self, filename=None, visible=False):
...
 
-    def close(self):
-        pass
+    def close(self, filepath = None):
```
elsevers: The MATLAB implementation is not taking in a filepath (that is, it does not save upon closing a document). Please coordinate with @ngadiyar93 to make your implementations consistent.

python/emach/tools/magnet/magnet.py
```diff
@@ -89,8 +100,15 @@ def draw_arc(self, centrexy, startxy, endxy):
         ret : TokenDraw object
 
         """
+        centre_x = eval(self.default_length)(centrexy[0])
+        centre_y = eval(self.default_length)(centrexy[1])
```
elsevers: Let's use american style spelling:

`centre_x` should be `center_x`
`centrexy` should be `center_xy`

python/emach/tools/magnet/magnet.py
```diff
@@ -115,9 +133,12 @@ def prepare_section(self, inner_coord):
         None.
 
         """
+        inner_coord_x = eval(self.default_length)(inner_coord[0])
```
elsevers: This function should have `csToken` as its argument, not `inner_coord`.

`inner_coord` should be a field of `csToken`. Please make this consistent with the MATLAB implementation.

python/emach/tools/magnet/magnet.py
```diff
@@ -115,9 +133,12 @@ def prepare_section(self, inner_coord):
...
+        inner_coord_y = eval(self.default_length)(inner_coord[1])
+        
         self.view.selectAt(
-            inner_coord[0],
-            inner_coord[1],
+            inner_coord_x,
+            inner_coord_y,
             self.consts.infoSetSelection,
             [self.consts.infoSliceSurface],
         )
```
elsevers, line 157: The function should also return an integer that is used in resulting token. I know that MagNet doesn't need that for its sake, but other tools do. 

If needed, please review the flow of information in MATLAB. I remember typing it up as a comment in a review on one of your previous PRs

python/emach/tools/magnet/magnet.py
```diff
@@ -155,4 +182,42 @@ def revolve(self, name, material, center, axis, angle, token = None):
         return ret
     
     def view_all(self):
+        '''
+        Function to view entire cross-section of drawing made in MagNer
+
+        '''
         self.view.viewAll()
+        
+    def set_default_length(self, user_unit, make_app_default):
```
elsevers, line 206: See the comment I just gave to @ngadiyar93 on PR #186 . Let's update this function so that `user_unit` takes in eMach vocabulary (`DimMillimeter`, `DimInch`, etc.)

Also see my comment to @ngadiyar93 about adding a `set_default_angular` function

elsevers (changes requested):

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: This has been corrected

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: The Matlab implementation has been replicated in the latest commit

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: This does not cause an error even if the document does not have a solution.

python/emach/tools/magnet/magnet.py
```diff
@@ -4,43 +4,62 @@
...
         else:
-            self.doc = self.disp_ex.newDocument()
+            self.doc = self.mn.newDocument()
         self.view = self.doc.getView()
         self.sol = self.doc.getSolution()
-        self.consts = self.disp_ex.getConstants()  # Get MAGNET Constants
-
+        self.consts = self.mn.getConstants()  # Get MAGNET Constants
+        
+        self.set_default_length(length_unit, False)
```
ASHAD1, line 48: Are you setting the default angle like what you have for length(`set_default_length`)? I am not sure about this in MagNet. I see that you have the property `default_angle`.

python/emach/tools/magnet/magnet.py
```diff
@@ -4,43 +4,62 @@
...
         self.consts = None
+        self.default_length = None
+        self.default_angle = None
+        self.filename = None
+        
+    def __del__(self):
+        self.mn.close(False)
 
-    def open(self, filename=None, visible=False):
+    def open(self, filename=None, length_unit = 'DimMillimeter', angle_unit = 'DimDegree'):
```
ASHAD1, line 32: Do we need to specify the length and angle unit here? I see that you are using `DimMillimeter` and `DimDegree` as your default length and angle. I think its better to specify the default unit in `_init` (`default_length` and `default_angle`) itself. This will make the `open()` signature similar to Matlab implementation.

python/emach/tools/magnet/magnet.py
```diff
@@ -63,12 +82,17 @@ def draw_line(self, startxy, endxy):
...
-        self.disp_ex.setVariant(0, 'line', 'python')
-        line = self.disp_ex.getVariant(0, 'python');
+        self.mn.setVariant(0, 'line', 'python')
+        line = self.mn.getVariant(0, 'python');
         
-        return TokenDraw(line,1)
+        return TokenDraw(line,0)
 
 
     def draw_arc(self, centrexy, startxy, endxy):
```
elsevers: rename `centrexy` to be `centerxy`

elsevers (changes requested):
@Bharat-Ramadas, nice job. One very minor request on a function argument name. Revise this and I will approve.

python/emach/tools/magnet/magnet.py
Bharat-Ramadas: Done!

elsevers (approved):
Looks good.

aimthiazz (approved):

ASHAD1 (approved):

PR #186, opened 2021-03-10
Update MagNet MATLAB tool

### LINKED ISSUE #172: Tools should use open and close for files, not the application instance

ngadiyar93:
This PR is does the following:

1. Update MATLAB MagNet tool to separate responsibility between `open ( )` and the constructor function. 
   a. The constructor creates a MagNet instance and sets its visibility.
   b. The `open( )` function only opens a MagNet document.
   
2. Implement `save ( )` function to save the drawn / solved MagNet model. Ideally before closing, but can be used at any point.

3. Implement `close ( )` function to close the MagNet document

4. Implement the destructor function  `delete ( )` function to close the MagNet instance and destroy the object.
   
5. Update `exampleParallelogram` to demonstrate the workflow.

6. Ignore files with extensions `.mn` - These are MagNet FEA files that we typically do not want in the repo.

The workflow is as follows:
`constructor` -- > `open( )` --> (Use plumbing / porcelain functions) --> `save( )` -- > `close( )` --> `delete( )`  

Updates:
1. This PR also updates the  `ToolBase` by adding `save`, `saveAs`, and additional arguments to `open`

2. Examples have been updated to use the new function signatures.

### DIFF
--- a/tools/ToolBase.m
+++ b/tools/ToolBase.m
@@ -13,7 +13,7 @@ classdef ToolBase < handle
     end
     
     methods(Abstract = true)
-        open(obj, fileName);
+        open(obj, path, fileName);
         save(obj, path, fileName)
         close(obj);
     end
--- a/tools/ToolBase.m
+++ b/tools/ToolBase.m
@@ -14,7 +14,8 @@ classdef ToolBase < handle
     
     methods(Abstract = true)
         open(obj, fileName);
-        save(obj, fileName);
+        saveAs(obj, fileName);
+        save(obj);
         close(obj);
     end
     
--- a/tools/ToolBase.m
+++ b/tools/ToolBase.m
@@ -13,8 +13,9 @@ classdef ToolBase < handle
     end
     
     methods(Abstract = true)
-        open(obj);
-        close(obj);
+        open(obj, fileName);
+        save(obj, path, fileName)
+        close(obj, fileName);
     end
     
      methods(Access = protected)
--- a/tools/ToolBase.m
+++ b/tools/ToolBase.m
@@ -15,7 +15,7 @@ classdef ToolBase < handle
     methods(Abstract = true)
         open(obj, fileName);
         save(obj, path, fileName)
-        close(obj, fileName);
+        close(obj);
     end
     
      methods(Access = protected)
--- a/tools/ToolBase.m
+++ b/tools/ToolBase.m
@@ -13,8 +13,8 @@ classdef ToolBase < handle
     end
     
     methods(Abstract = true)
-        open(obj, path, fileName);
-        save(obj, path, fileName)
+        open(obj, fileName);
+        save(obj, fileName);
         close(obj);
     end
     
--- a/tools/ToolBase.m
+++ b/tools/ToolBase.m
@@ -13,10 +13,10 @@ classdef ToolBase < handle
     end
     
     methods(Abstract = true)
-        open(obj, fileName);
-        saveAs(obj, fileName);
-        save(obj);
-        close(obj);
+        open(obj, fileName); % Open a file
+        save(obj); % Save the current file
+        saveAs(obj, fileName); % Save the current file with the specified name at the specified location
+        close(obj); % Close the file
     end
     
      methods(Access = protected)
--- a/tools/magnet/MagNet.m
+++ b/tools/magnet/MagNet.m
@@ -8,35 +8,21 @@ classdef MagNet < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
     properties (GetAccess = 'public', SetAccess = 'private')
         mn;  % MagNet activexserver object
         doc; % Document object
-        view; %View object
+        view; % View object
         consts; %Program constants
         defaultLength = 'DimMillimeter';
         defaultAngle  = 'DimDegree';
+        visible = 1;% Default visibility
     end
     
     methods
         function obj = MagNet(varargin)
             obj = obj.createProps(nargin,varargin);            
-            obj.validateProps();            
+            obj.validateProps();        
+            obj.mn = actxserver('MagNet.Application');
+            set(obj.mn, 'Visible', obj.visible);
         end
-        
-        
-        function obj = create(obj, visible)
-            %CREATE Create a new MagNet instance 
-            %   create() creates a new instance of MagNet.
-            %
-            %   create(visible) sets the visibility of the created instance.
-            %   visible = 1 or 'true' for MagNet instance to be visible.
-            %   visible = 0 or 'false' for MagNet instace to be invisible.
-           obj.mn = actxserver('MagNet.Application');
-           if nargin > 1 && (visible  == 1 || visible == 0 ||...
-                   visible == 'true' || visible == 'false')
-              set(obj.mn, 'Visible', visible);
-           else
-              set(obj.mn, 'Visible', 1);
-           end
-        end
-
+                
          function open(obj, fileName)
             %OPEN Open MagNet
             %   open() opens a new instance of MagNet with a new document.
--- a/tools/magnet/MagNet.m
+++ b/tools/magnet/MagNet.m
@@ -26,10 +26,16 @@ classdef MagNet < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
          function open(obj, fileName)
             %OPEN Open MagNet
             %   open() opens a new instance of MagNet with a new document.
-            obj.doc = invoke(obj.mn, 'newDocument');          
-            obj.view = invoke(obj.doc, 'getview');
-            obj.consts = invoke(obj.mn, 'getConstants');
-            obj.setDefaultLengthUnit('millimeters', false);
+            %   open(fileName) opens an existing MagNet document. fileName
+            %   is a string that has the full filepath and the file name.
+            if isempty(fileName)
+                obj.doc = invoke(obj.mn, 'newDocument');
+            else
+                obj.doc = invoke(obj.mn, 'openDocument', fileName);
+            end
+                obj.view = invoke(obj.doc, 'getview');
+                obj.consts = invoke(obj.mn, 'getConstants');
+                obj.setDefaultLengthUnit('millimeters', false);
          end
         
          function save(obj, path, fileName)
--- a/tools/magnet/MagNet.m
+++ b/tools/magnet/MagNet.m
@@ -20,32 +20,28 @@ classdef MagNet < ToolBase & DrawerBase & MakerExtrudeBase & MakerRevolveBase
             obj.validateProps();            
         end
         
-        function obj = open(obj, fileName, mn, visible)
-            %OPEN Open MagNet or a specific file.
+        
+        function obj = create(obj, visible)
+            %CREATE Create a new MagNet instance 
+            %   create() creates a new instance of MagNet.
+            %
+            %   create(visible) sets the visibility of the created instance.
+            %   visible = 1 or 'true' for MagNet instance to be visible.
+            %   visible = 0 or 'false' for MagNet instace to be invisible.
+           obj.mn = actxserver('MagNet.Application');
+           if nargin > 1 && (visible  == 1 || visible == 0 ||...
+                   visible == 'true' || visible == 'false')
+              set(obj.mn, 'Visible', visible);
+           else
+              set(obj.mn, 'Visible', 1);
+           end
+        end
+
+         function open(obj, fileName)
+            %OPEN Open MagNet
             %   open() opens a new instance of MagNet with a new document.
             %
             %   open('filename') opens the file in a new instance of MagNet.
-            %
-            %   open('filename', mn) opens the file in the mn MagNet instance
-            %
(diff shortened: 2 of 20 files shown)


Bharat-Ramadas (dismissed):
Modified example works

Bharat-Ramadas (commented):
It just occurred to me that `visible` should not be a `protected` property. Users should be able to change visible whenever they wish to do so

ngadiyar93:
> It just occurred to me that `visible` should not be a `protected` property. Users should be able to change visible whenever they wish to do so

@Bharat-Ramadas Users can update visibility using `setVisibility( );` I just verified that this works. So there should not be an issue.

tools/magnet/MagNet.m
```diff
@@ -55,14 +37,23 @@
...
+                     sprintf('Call getDocument().save("%s")',filePath)); 
+         end
+         
         function close(obj)
-           % CLOSE Closes MagNet application
-           %     close()
-           
-           % TODO:
-           % Implement this...
+           % CLOSE Closes the MagNet document
```
ASHAD1, line 63: Here `close()` closes the document. Do we need to specify which document to close; should we pass `fileName` as an argument for the close function like we pass in the `open()` function?

tools/magnet/MagNet.m
```diff
@@ -8,44 +8,26 @@
...
-            obj.validateProps();            
+            obj.validateProps();        
+            obj.mn = actxserver('MagNet.Application');
+            set(obj.mn, 'Visible', obj.defaultVisibility);
         end
-        
-        function obj = open(obj, fileName, mn, visible)
-            %OPEN Open MagNet or a specific file.
+                
+         function open(obj, fileName)
```
elsevers, line 28: Change function so that the only path to creating a new document is to not have the filename argument populated. (So, don't support the `filename = 0` argument)

tools/magnet/MagNet.m
```diff
@@ -55,14 +37,23 @@
             obj.view = invoke(obj.doc, 'getview');
             obj.consts = invoke(obj.mn, 'getConstants');
             obj.setDefaultLengthUnit('millimeters', false);
-        end
+         end
         
+         function save(obj, path, fileName)
```
elsevers: Need to add this function to the abstract base class `ToolBase`

tools/magnet/MagNet.m
elsevers, line 28: `ToolBase` abstract base class needs to have the correct function signature

Question: do we need to specify each viable function signature in `ToolBase`?

tools/magnet/MagNet.m
elsevers: Let's have a policy that we do not open multiple documents with the same eMach tool object.

elsevers (changes requested):

tools/magnet/MagNet.m
```diff
@@ -8,44 +8,26 @@
     properties (GetAccess = 'public', SetAccess = 'private')
         mn;  % MagNet activexserver object
         doc; % Document object
-        view; %View object
+        view; % View object
         consts; %Program constants
         defaultLength = 'DimMillimeter';
-        defaultAngle  = 'DimDegree'
+        defaultAngle  = 'DimDegree';
+        defaultVisibility = 0;% Default visibility
```
elsevers: This property name of `defaultVisibility` doesn't make sense now that we have revised the class `open` and `close` methods. Let's rename it to just be `visible`. Please also update the function `setVisibility` to update this variable.

tools/magnet/MagNet.m
```diff
@@ -8,44 +8,26 @@
...
     end
     
     methods
         function obj = MagNet(varargin)
             obj = obj.createProps(nargin,varargin);            
-            obj.validateProps();            
+            obj.validateProps();        
+            obj.mn = actxserver('MagNet.Application');
+            set(obj.mn, 'Visible', obj.defaultVisibility);
```
elsevers: See comment above about `visible`

tools/magnet/MagNet.m
```diff
@@ -55,14 +37,23 @@
             obj.view = invoke(obj.doc, 'getview');
             obj.consts = invoke(obj.mn, 'getConstants');
             obj.setDefaultLengthUnit('millimeters', false);
```
elsevers:
```matlab
obj.setDefaultLengthUnit('millimeters', false);
```

This needs to be set based on the value of `defaultLength`

tools/magnet/MagNet.m
elsevers: This function is behaving as a `saveAs` function -- that is, it is taking in a file name that should be written out. Seems like we should also support a `save` function, where MagNet already knows the path it should save to. Does `MagNet` have such a function in its API? Or would we need to be tracking the file path in our tool class from the last save?

elsevers (changes requested):
We still have a few issues with units here. I know that I previously said it is okay for `setDefaultLengthUnit` to take in a unit in MagNet language vocabulary (`millimeters`, `inches`, etc.), but I no longer think this makes sense.

https://github.com/Severson-Group/eMach/blob/e2d5828317a64945588afdda46dd9e86ed4c91a2/tools/magnet/MagNet.m#L168-L197

Instead, this function should take in units in `eMach` vocabulary and then set `MagNet` to the appropriate unit. For example, the uses passes in `userUnit = 'DimMillimeter'` and then this function calls `invoke(obj.doc, 'setDefaultLengthUnit', 'millimeters', makeAppDefault); `

If you do this, then you should use this function at the end of your constructor to get MagNet using the correct units that match your `defaultLength` and `defaultAngle`.

Note that this means that you also need to add a function `setDefaultAngularUnit` (or does MagNet only support degrees?)

ngadiyar93:
> We still have a few issues with units here. I know that I previously said it is okay for `setDefaultLengthUnit` to take in a unit in MagNet language vocabulary (`millimeters`, `inches`, etc.), but I no longer think this makes sense.
> 
> https://github.com/Severson-Group/eMach/blob/e2d5828317a64945588afdda46dd9e86ed4c91a2/tools/magnet/MagNet.m#L168-L197
> 
> Instead, this function should take in units in `eMach` vocabulary and then set `MagNet` to the appropriate unit. For example, the uses passes in `userUnit = 'DimMillimeter'` and then this function calls `invoke(obj.doc, 'setDefaultLengthUnit', 'millimeters', makeAppDefault); `
> 
> If you do this, then you should use this function at the end of your constructor to get MagNet using the correct units that match your `defaultLength` and `defaultAngle`.
> 
> Note that this means that you also need to add a function `setDefaultAngularUnit` (or does MagNet only support degrees?)

@elsevers I have made this update to `setDefaultLength`. 
Yes, MagNet only supports degrees for angles. It does not support radians....so there is no function to set default angular unit in MagNet.

tools/magnet/MagNet.m
ngadiyar93: @elsevers I have made the update to the `ToolBase`. I am not very sure if what I have done is exactly what is expected. Please let me know... Also, Matlab doesnt seem to require that every viable signature is specified. I checked this by opening MagNet with different number of arguments and it didnt break.

tools/magnet/MagNet.m
ngadiyar93: I checked this -- Magnet only has a save function that requires the filename + path to be specified. 
Even if an existing document that has already been saved earlier is re-saved in the MagNet gui by hitting `Save `(not `Save As`),  The session log still records the following:
image
I think this means we need to track the path from the last save in the tool class.

tools/ToolBase.m
```diff
@@ -13,7 +13,8 @@
     end
     
     methods(Abstract = true)
-        open(obj);
+        open(obj, fileName);
```
ASHAD1: For `open()`, do we also need to specify the path of the file to open?

ngadiyar93: Updated to specify the path in the function signature.

Bharat-Ramadas (dismissed):
Looks good

ASHAD1 (dismissed):

tools/ToolBase.m
```diff
@@ -13,7 +13,8 @@
     end
     
     methods(Abstract = true)
-        open(obj);
+        open(obj, path, fileName);
```
elsevers: Why should the user have to specify both `path` and `filename`? I like the function signature you previously provided in the last round in `MagNet.m`:
`open(obj, fileName)` 

Let's go back to this.

tools/magnet/MagNet.m
```diff
@@ -8,61 +8,60 @@
...
-                    obj.mn = actxserver('MagNet.Application');
-                end
                 
-                set(obj.mn, 'Visible', visible);
-            end
-
-            if nargin > 0 && ~isnumeric(fileName)
-                obj.doc = invoke(obj.mn, 'openDocument', fileName);
-            else
+         function open(obj, varargin)
```
elsevers: 1) It doesn't seem to like a good use of `varargin` if there is an optional function argument and 
2) this does not match your tool base class. 

Please use this function signature:
`open(obj, fileName)`

tools/ToolBase.m
```diff
@@ -13,7 +13,8 @@
...
+        save(obj, path, fileName)
```
elsevers: Again, remove `path`

tools/magnet/MagNet.m
elsevers: Great, can you implement that?

Also, remove `path` from the function signature. Too complicated for a user.

elsevers (changes requested):
Improvements look good. I have added a few requests as inline comments.

This code breaks nearly all examples and cannot be merged until this is fixed.

tools/magnet/MagNet.m
ngadiyar93: @elsevers Question on this.
The `save` function where we track the path is only applicable if a user opens an existing file on the disk by specifying `fileName` in the `open` function. i.e, it is applicable only if the user calls `open(fileName)` by specifying an existing fileName and not simply `open( )`. 
Is that right? because I cannot see how one can `save` an untitled document opened by calling `open( )` with no path specified -- it has to be a `save as` for that case.

elsevers: Correct. The class needs to through an error if the objects internal filename variable has not been set.

ngadiyar93: Nevermind, i think I figured this out.

tools/magnet/MagNet.m
```diff
@@ -213,11 +210,12 @@ function viewAll(obj)
...
-        function setVisibility(obj, visibility)
+        function setVisibility(obj, visible)
             %SETVISIBLE Sets visibility of MagNet application
-            %    setVisibility(true)
-            
-            set(obj.mn, 'Visible', visibility);
+            %    Set visible= 'true' or visible = 1 to make the MagNet
+            %    session visible. Set visible= 'false' or visible = 0 to 
+            %    make the MagNet session invisible.  
+            set(obj.mn, 'Visible', visible);
```
elsevers, line 217: If we are tracking `visible` as an object property (and we are), we better update it here when this function is called.

tools/magnet/MagNet.m
```diff
@@ -8,61 +8,66 @@
...
+                obj.consts = invoke(obj.mn, 'getConstants');
+                obj.setDefaultLengthUnit(obj.defaultLength, false);
+         end
+        
+         function saveAs(obj, fileName)
+            % SAVEAS Save the MagNet document.
+            % fileName is a string that specifies the complete path to the file.
+            invoke(obj.mn, 'processcommand',...
+                     sprintf('Call getDocument().save("%s")',fileName));
+            obj.fileName = fileName;
```
elsevers, line 47: In an effort to make the code more DRY, I suggest revising this function to be:

```matlab
obj.fileName = fileName;
obj.save()
```

elsevers (changes requested):
I think this is just about ready to merge. Just two small change requests.

ngadiyar93:
@elsevers I have addressed your change requests.

elsevers (approved):
Looks good. Thanks @ngadiyar93!

Bharat-Ramadas (approved):
Looks good

PR #194, opened 2021-03-20
Update breadloaf cross-section

Bharat-Ramadas:
# Description
This PR makes the `breadloaf` cross-sect compliant with the format specified for cross-sects in `eMach`.


aimthiazz (approved):
I thought we need a PDF document along with SVG file. However, I do not see PDF requirements in the crossect markdown.  Other than that, the drawing looks good.

elsevers (approved):
Looks good.

PR #209, opened 2021-04-19
Add function to union components in MagNet

ngadiyar93:
This PR adds a function to union two or more components into a single component in MagNet. This is particularly useful for Ryan's  conical bearing modeling project.


elsevers (approved):
Thanks @ngadiyar93. I have not tested this, but the code looks reasonable to me.

Bharat-Ramadas (approved):

PR #226, opened 2021-11-02
Add JMAG Python tool to eMach

Bharat-Ramadas:
This PR implements the JMAG Python tool for eMach. During the implementation of this tool, a bug was observed in the `Location2D` class which caused the drawing of components to take on incorrect dimensions within the tool. This bug will be addressed by this PR. Changes have been made in each of the cross-section classes to reflect the updates made in `Location2D`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ octave-workspace
 *.pyc
 __pycache__/
 .idea/workspace.xml
+.idea/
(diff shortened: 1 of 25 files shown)


elsevers (approved):
Thanks @Bharat-Ramadas  and @aimthiazz

aimthiazz (approved):

PR #227, opened 2021-11-30
Add functionality for JMAG 2D analysis

Bharat-Ramadas:
This PR does the following:

1. Modifies the JMAG tool to create `sketches` instead of `parts` when drawing cross-sections as `parts` are inherently 3D in JMAG and `sketches` are the 2D `eMach` cross-section equivalent. 
2. Moves part creation over to the `extrude` and `revolve` methods as this is where 3D components are created.
3. Implements a new `JmagDesigner2D` tool for 2D analysis with JMAG. This class reuses most of the `JmagDesigner` tool methods, with only the `extrude` and `revolve` functions modified. This is required as JMAG determines the type of analysis to be 2D or 3D based on the geometry.


elsevers (changes requested):
There seem to be a lot of "white-space only" changes in here. Are these intentional? 

If they are not, let's back out the changes. It is not a good practice to commit white space changes were inadvertently created (okay to commit them if you are doing this for a purpose, such as cleaning up a file)

Bharat-Ramadas:
The added whitespaces are intentional. These changes make the files more PEP compliant (mainly [PEP 8](https://www.python.org/dev/peps/pep-0008/)) in terms of spacing and other style guidelines.

elsevers:
Can you please add a description to this PR? What functionality are you adding?

aimthiazz (approved):

elsevers (approved):

PR #228, opened 2021-12-10
Add inner reluctance rotor cross section

Aneesh-Jategaonkar:
This PR adds inner reluctance rotor cross section drawing capability to the Python implementation of eMach. The example script corresponding to this cross-section is python/examples/example_inner_reluctance_rotor.py


Bharat-Ramadas (commented):
Nice work @Aneesh-Jategaonkar
I would suggest the following changes:
1. Change the class docustring to reflect the details related to your cross-section
2.  Include a markdown document that will enable viewing the cross-section within Git. This [link](https://github.com/Severson-Group/eMach/tree/add-python-support/python/emach/model_obj/cross_sects/notched_rectangle) illustrates what I mean by this comment.

elsevers (changes requested):
Thanks Aneesh. However, this PR seems to be missing the `README.md` file necessary to be compliant with eMach cross sections. The image of the cross-section also is not using the naming convention.

Please review eMach's guidelines for creating new cross-sections [here](https://github.com/Severson-Group/eMach/tree/develop/model_obj/crosssects).

Bharat-Ramadas (commented):
Looks good to be merged in! 
Only comment is that the fillet feature has not been incorporated into the cross-section. Although this isn't critical, it might be a good idea to revise the cross-section image to remove the fillets so that there is no confusion moving forward.

elsevers:
> Looks good to be merged in! Only comment is that the fillet feature has not been incorporated into the cross-section. Although this isn't critical, it might be a good idea to revise the cross-section image to remove the fillets so that there is no confusion moving forward.

Actually, the fillet feature is quite important because we are planning to do structural analysis on this cross-sect. Without being able to set fillets, we'll likely have stress concentrations. So please do add the fillet feature.

elsevers (changes requested):
- [x] When I change your example to use `P=5`, JMAG crashes --see the error below. Please fix.
image

- [x] Please make `p` be lower case.
- [x] Please implement fillets

Bharat-Ramadas (dismissed):
Looks good!

elsevers (changes requested):
This is getting very close. 

- [x] Update dimensions. Please take a look at the [inner notched rotor cross-section](https://github.com/Severson-Group/eMach/tree/Add_synrel_rotor_cross-section/python/emach/model_obj/cross_sects/inner_notched_rotor) and make your dimensions reasonably consistent with this:
    1. Change your `r_sh` to be `r_ri`
    2. Instead of `r_ri` use a yoke thickness `d_ri`
    3. Instead of `r_ro` use a pole depth of `d_rp`
- [x] Add a pole pitch annotation or change `p = number of pole pairs` to be `p = number of teeth`
- [x] There is no reason that this cross-section is only a synchronous reluctance rotor. It could be a switched reluctance rotor. Or, someone could even put permanent magnets between the teeth and make it a consequent pole rotor. We also have been specific in our rotor/stator naming of specifying whether it is for inner or outer. For all these reasons, let's rename this cross-section to be `Inner Reluctance Rotor`.

Bharat-Ramadas (commented):
Looks great! Only recommendation I have is to change `p` to `Q_r` so that users don't confuse this variable for number of pole pairs (Q_r for rotor teeth). Also note that this will have to be changed in the .svg file. Apart from that, this PR looks ready to be merged in! Nice job!

Aneesh-Jategaonkar:
> Looks great! Only recommendation I have is to change `p` to `Q_r` so that users don't confuse this variable for number of pole pairs (Q_r for rotor teeth). Also note that this will have to be changed in the .svg file. Apart from that, this PR looks ready to be merged in! Nice job!

Variable name p is changed to Q_r and the .svg file is also updated

Bharat-Ramadas (approved):
Looks good!

elsevers (approved):
Nice job @Aneesh-Jategaonkar

PR #229, opened 2021-12-10
Add flux barrier rotor cross section

Aneesh-Jategaonkar:
This PR adds the flux barrier rotor cross section drawing capability to the Python implementation of eMach. The example script corresponding to this cross-section is python/examples/example_flux_barrier_rotor.py


elsevers (changes requested):
Thanks Aneesh. However, this PR seems to be missing the `README.md` file necessary to be compliant with eMach cross sections. The image of the cross-section also is not using the naming convention.

Please review eMach cross-section guidelines [here](https://github.com/Severson-Group/eMach/tree/develop/model_obj/crosssects).

Bharat-Ramadas (commented):
The cross-section looks great! Nice job!
Would suggest the following changes:

1. Right now it seems that the number the flux barriers per pole is fixed at 3. It might be worthwhile to make this a variable as well.
2. Remove fillets in the cross-section figure as this is currently not implemented.

elsevers (changes requested):
Looking good. I am confused about the dimensions `l_m1`, `l_m2`, `l_m3`. It seems to me that these dimensions are tied only to magnets and have nothing to do with the dimensions of the IPM rotor cross-section. Why are they needed for this class?

There are many types of IPM rotor cross-sections. Is there a more descriptive name that we can be using for this type of IPM rotor?

Does this class always need to have 3 magnets / bridge sections per pole? It looks like that is non-configurable (that's okay if it is), is that correct?

elsevers:
> The cross-section looks great! Nice job! Would suggest the following changes:
> 
>     1. Right now it seems that the number the flux barriers per pole is fixed at 3. It might be worthwhile to make this a variable as well.
> 
>     2. Remove fillets in the cross-section figure as this is currently not implemented.

1. It would be wonderful if this was a variable.
2. Fillets are important! Please implement them.

Aneesh-Jategaonkar:
> Looking good. I am confused about the dimensions `l_m1`, `l_m2`, `l_m3`. It seems to me that these dimensions are tied only to magnets and have nothing to do with the dimensions of the IPM rotor cross-section. Why are they needed for this class?
> 
> There are many types of IPM rotor cross-sections. Is there a more descriptive name that we can be using for this type of IPM rotor?
> 
> Does this class always need to have 3 magnets / bridge sections per pole? It looks like that is non-configurable (that's okay if it is), is that correct?

l_m1, l_m2 and l_m3 dimensions were deleted. They are not needed for this class. I have changed the name to Flux Barrier Rotor. The bridge sections per pole are presently non-configurable. Fillets are added.

Bharat-Ramadas (commented):
Looking good. Only changes I would suggest is to switch from `P = number of poles` to `p = number of pole pairs`, for consistency with other cross-sections

Aneesh-Jategaonkar:
> Looking good. Only changes I would suggest is to switch from `P = number of poles` to `p = number of pole pairs`, for consistency with other cross-sections

The change has been made and updated in the .svg file as well

Bharat-Ramadas (dismissed):
Looks good!

elsevers (dismissed):
Nice job @Aneesh-Jategaonkar

elsevers (approved):
Nice job @Aneesh-Jategaonkar

PR #230, opened 2021-12-12
Add capability to draw and create coil conditions in JMAG2D

Bharat-Ramadas:
This PR implements the following:
1. A new cross-section to draw coil sides in 2D
2. New classes to represent machine windings (coils -> coil sides -> winding)
3. Methods implemented in JMAG2D to support creating winding conditions 

This PR should be merged in only after #227 is merged.


elsevers:
Since it sounds like we don't plan to merge this, I am closing this PR.

